PR #2, opened 2026-05-22
Linux support for exoharness

akrentsel:
## Summary

Makes exoharness work on Linux. Previously the harness hardcoded macOS-only dependencies (Apple Keychain for secret encryption, Apple's `container` CLI for sandboxes), so `exo secret set` failed immediately on any non-macOS host and the chat REPL couldn't get past harness construction.

updated the approach here to be static (gated by `cfg(target_os = "..."). 

### `exoharness: file-backed master key provider for Linux`

Adds `FileBackedSecretKeyProvider` mirroring the shape of `AppleKeychainSecretKeyProvider`: lazy `OnceLock` cache, read existing key on first access, or generate a random 32-byte key and persist it. The key file lives at `$XDG_CONFIG_HOME/exo/master.key` (or `$HOME/.config/exo/master.key`), written atomically via tmp+rename with file mode `0600` and parent directory `0700`. A cfg-gated `default_secret_cipher` is what the harness constructors call.

Noting the threat model here: this does not give us real encryption-at-rest, because an attacker with access to this machine can find the key and decrepyt the encrypted secrets. I looked at what some other tools do (aws cli, gh, etc.) and this seems to be acceptable to them. The one thing we do to provide some slight safety is we store the key in a different, global directory than the local secrets. so it is harder to accidentally leak. 

I structured the architecture this way on: swapping in a real keystore (1Password Connect, Vault, AWS Secrets Manager) later is a new `SecretKeyProvider` impl, no other code changes.

### `exoharness: docker sandbox backend for Linux`

Generalizes the Apple-only sandbox backend into `CliContainerSandboxBackend` parameterized by `ContainerCliFlavor`, with `apple_container()` and `docker()` constructors. The Apple flavor keeps every shell invocation byte-for-byte; the Docker flavor uses the same `run`/`exec` shape and substitutes `docker rm -f` for the Apple `stop`+`delete` two-step.

The `rm -f` substitution isn't cosmetic: `docker stop` defaults to a 10s SIGTERM grace, which races the harness's 5s `WARM_SANDBOX_CLEANUP_TIMEOUT` and leaves `Exited` containers behind that the subsequent `docker rm` can't remove. `rm -f` is SIGKILL + remove in one command; for sleep-infinity warm sandboxes there is no graceful state to preserve.

Apple-specific bits (system-start, orphan reaper, Apple-absolute-time deserializer) stay confined to the apple_container arm via `cli.requires_system_start()` and `match cli`. A cfg-gated `default_sandbox_backend` returns `apple_container()` on macOS and `docker()` on Linux. Adding a Podman backend later is a new `ContainerCliFlavor` variant + a new constructor.

## End-to-end verified on Linux

- `exo secret set <name> --env <ENV>` succeeds (previously failed at keychain init).
- `master.key` file written 0600, parent dir 0700, 32 bytes.
- Secrets round-trip across separate `exo` invocations (multi-process master-key read works).
- On-disk secret blobs are AES-256-GCM ciphertext with per-secret nonces (verified by inspecting `<root>/exoharness/secrets/<uuid>.json`).
- `exo chat repl <agent> <conversation>` against `gpt-4o-mini` completes turns successfully — secret decrypts via file-backed master key, sandbox container starts via Docker, streaming response arrives token-by-token.
- `docker rm -f` cleanup in `Drop for CliContainerSandboxBackend` leaves zero containers behind after the harness drops (previously each chat invocation would leak one `Exited` container under a `stop` + `rm` two-step).

## macOS verification

Not directly verified on macOS hardware. The macOS code paths are **structurally preserved**: every macOS-flavor invocation produces byte-identical shell commands to before this PR, and the cfg gates mean the macOS-specific code is identical to what's on `main`.

I'm going to look at adding some ci testing to make sure stuff works on both macos and linux. 

## Test plan

- [x] Linux end-to-end (chat repl, multi-turn, parallel tool calls — all work)
- [x] `cargo test -p exoharness --features basic-backend` (10/10 pass)
- [x] `cargo clippy -p exoharness --all-targets --features basic-backend` (no warnings)
- [ ] macOS verification via follow-up CI PR

### DIFF
--- a/crates/exoharness/src/basic.rs
+++ b/crates/exoharness/src/basic.rs
@@ -18,7 +18,7 @@ use crate::sandbox::{
     SandboxSpec,
 };
 use crate::secrets::{
-    AppleKeychainSecretKeyProvider, EncryptedSecret, SecretCipher, StaticSecretKeyProvider,
+    EncryptedSecret, SecretCipher, StaticSecretKeyProvider, default_secret_cipher,
 };
 use crate::storage::BasicObjectStore;
 use crate::{
@@ -50,9 +50,7 @@ impl BasicExoHarness {
     pub async fn new(root: impl Into<PathBuf>) -> Result<Self> {
         let root = root.into();
         let storage = BasicObjectStore::local_filesystem(&root).await?;
-        let secret_cipher = SecretCipher::new(Arc::new(AppleKeychainSecretKeyProvider::new(
-            root.to_string_lossy().to_string(),
-        )));
+        let secret_cipher = default_secret_cipher(root.to_string_lossy().to_string())?;
         Self::new_with_components(
             storage,
             secret_cipher,
@@ -73,10 +71,7 @@ impl BasicExoHarness {
     }
 
     pub async fn new_with_object_store(store: Arc<dyn object_store::ObjectStore>) -> Result<Self> {
-        let keychain_account = store.to_string();
-        let secret_cipher = SecretCipher::new(Arc::new(AppleKeychainSecretKeyProvider::new(
-            keychain_account,
-        )));
+        let secret_cipher = default_secret_cipher(store.to_string())?;
         Self::new_with_components(
             BasicObjectStore::new(store),
             secret_cipher,
@@ -89,10 +84,7 @@ impl BasicExoHarness {
         store: Arc<dyn object_store::ObjectStore>,
         sandbox_backend: Arc<dyn ManagedSandboxBackend>,
     ) -> Result<Self> {
-        let keychain_account = store.to_string();
-        let secret_cipher = SecretCipher::new(Arc::new(AppleKeychainSecretKeyProvider::new(
-            keychain_account,
-        )));
+        let secret_cipher = default_secret_cipher(store.to_string())?;
         Self::new_with_components(BasicObjectStore::new(store), secret_cipher, sandbox_backend)
             .await
     }
--- a/crates/exoharness/src/basic.rs
+++ b/crates/exoharness/src/basic.rs
@@ -12,10 +12,10 @@ use tokio::sync::{Mutex as AsyncMutex, mpsc};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 
 use crate::sandbox::{
-    AppleContainerSandboxBackend, LocalProcessSandboxBackend, ManagedSandboxBackend,
-    ManagedSandboxHandle, SANDBOX_MAIN_MOUNT_DIR, SandboxCommand, SandboxKey,
-    SandboxLifecycleConfig, SandboxMount, SandboxMountAccess, SandboxNetworkPolicy, SandboxRequest,
-    SandboxSpec,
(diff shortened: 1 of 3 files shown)


crates/exoharness/src/platform.rs
```diff
@@ -0,0 +1,30 @@
+use std::env;
+
+use anyhow::bail;
+
+use crate::Result;
+
+const PLATFORM_ENV_VAR: &str = "EXO_PLATFORM";
```
ankrgyl: prefer using clap for these arguments (so they can be provided via env or cli)

crates/exoharness/src/secrets.rs
```diff
@@ -167,3 +208,71 @@ fn ensure_apple_keychain_store() -> Result<()> {
...
+pub(crate) fn secret_cipher_for_platform(
+    platform: Platform,
+    keychain_account: String,
+) -> Result<SecretCipher> {
+    let provider: Arc<dyn SecretKeyProvider> = match platform {
+        Platform::MacOs => Arc::new(AppleKeychainSecretKeyProvider::new(keychain_account)),
+        Platform::Linux => Arc::new(FileBackedSecretKeyProvider::new(default_master_key_path()?)),
+    };
+    Ok(SecretCipher::new(provider))
+}
```
ankrgyl: what is the use case for this to be dynamic? i guess i could test what it would be like to use linux on a mac (because right now it's really only secret management that's different?)

akrentsel: yeah i think you're right, this doesn't need to be dynamic, I started down the path of thinking you may want to use the linux behavior on mac, but for that we can add support in the future of explicitly choosing the container backend or secret configuration. not needed for now.

crates/exoharness/src/platform.rs
akrentsel: no longer necessary since it's not dynamic

akrentsel:
Running on linux, with docker containers & local secrets! 
image

crates/exoharness/src/basic.rs
```diff
@@ -50,13 +50,11 @@ impl BasicExoHarness {
...
         let storage = BasicObjectStore::local_filesystem(&root).await?;
-        let secret_cipher = SecretCipher::new(Arc::new(AppleKeychainSecretKeyProvider::new(
-            root.to_string_lossy().to_string(),
-        )));
+        let secret_cipher = default_secret_cipher(root.to_string_lossy().to_string())?;
         Self::new_with_components(
             storage,
             secret_cipher,
-            Arc::new(AppleContainerSandboxBackend::new()),
+            default_sandbox_backend(),
```
ankrgyl, line 57: i think sandbox backend should be an argument so if you're on mac os x you can use docker. Also, we'll eventually (soon!) want to support remote sandboxes and make the experience seamless

akrentsel:
Superseded by #9, which takes the alternate approach of making secret and sandbox backends user-selectable through BasicExoHarnessConfig rather than cfg-gating on target_os. Linux support is delivered the same way (file-backed secrets, Docker sandbox), but without compile-time platform branching.

PR #3, opened 2026-05-22
Make it trivial to get to a REPL

Alexsun1one:
## What

`exo repl` makes getting to a chat a single command. After registering a model,
it reuses or creates a default agent and conversation and uses an existing model
binding — no `agent create` / `conversation create` / `chat repl` dance.

Reworked per review:
- It does not read credentials from the environment. Registering a model stays
  explicit (`exo secret set` / `exo model register`); `exo repl` just uses what's
  registered and errors with guidance if nothing is.
- The default conversation has no shell tool, so no sandbox is provisioned — it's
  a plain chat. Create a regular conversation when you want tools. No exoharness
  changes.

## Test plan

- `cargo test --workspace --all-targets` (added: `exo repl` parses with no args /
  with `--model`; resolves to an existing model binding)
- `cargo clippy --workspace --all-targets` — no new warnings

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -333,22 +333,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             conversation,
         } => {
             let agent_slug = agent.unwrap_or_else(|| DEFAULT_REPL_SLUG.to_string());
-            let conversation_slug = conversation.unwrap_or_else(|| DEFAULT_REPL_SLUG.to_string());
+            // Without --conversation, start a fresh session each run (the usual CLI
+            // behavior); pass --conversation <slug> to resume or target a specific one.
+            let conversation_slug = conversation.unwrap_or_else(generate_fun_slug);
 
-            let existing_agent = harness.get_agent(&agent_slug).await?;
-            if existing_agent.is_some() && model.is_some() {
-                eprintln!(
-                    "note: agent '{agent_slug}' already exists; --model only applies when creating it"
-                );
-            }
-            let agent = match existing_agent {
+            let agent = match harness.get_agent(&agent_slug).await? {
                 Some(agent) => agent,
                 None => {
                     let model = ensure_repl_model(harness.as_ref(), model).await?;
-                    let agent = harness
+                    harness
                         .create_agent(CreateAgentRequest {
-                            slug: agent_slug,
-                            name: Some("REPL".to_string()),
+                            slug: agent_slug.clone(),
+                            name: Some(agent_slug),
                             harness: to_agent_harness_kind(harness_kind),
                             typescript: None,
                             sandbox_image: None,
@@ -358,13 +354,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                             max_tool_round_trips: None,
                             braintrust: None,
                         })
-                        .await?;
-                    println!(
-                        "created agent {} ({})",
-                        agent.record().slug,
-                        agent.record().id
-                    );
-                    agent
+                        .await?
                 }
             };
 
@@ -373,8 +363,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 None => {
                     let conversation = agent
                         .create_conversation(CreateConversationRequest {
-                            slug: Some(conversation_slug),
-                            name: Some("REPL".to_string()),
+                            slug: Some(conversation_slug.clone()),
+                            name: Some(conversation_slug),
                         })
                         .await?;
                     // The default REPL is a plain chat: drop the shell tool so no
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -6,6 +6,8 @@ mod mount_tests;
 #[cfg(test)]
 mod naming_tests;
 #[cfg(test)]
+mod repl_tests;
+#[cfg(test)]
 mod secret_tests;
 mod tui;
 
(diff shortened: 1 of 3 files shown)


crates/exoharness/src/basic.rs
```diff
@@ -1464,11 +1449,50 @@ impl TurnHandle for BasicTurnHandle {
...
+        Ok(add_result)
     }
+
+    /// Rebases only across exoharness-managed sandbox lifecycle events recorded
+    /// while this turn was running; any other intervening event is a genuine
+    /// concurrent write and stays a head mismatch.
+    async fn reconcile_head(
+        &self,
+        turn_head: Option<EventId>,
+        conversation_head: Option<EventId>,
```
ankrgyl: instead of this logic, i think it would be cleaner for the `exo repl` default to _not_ have a sandbox available.

ankrgyl (commented):
Hi! Thanks for the contribution!

I'm generally a fan of the idea of getting you into a repl (with the built-in default harness) w/o a sandbox, but I do think it is worth splitting the secrets commands out because the exoharness is meant to be a "no frills" and secure system, and default fetching something from your environment is a bit sketchy. Also, a lot of more security conscious folks won't have `OPENAI_API_KEY` set in their environment, and will miss this.

Can you rework the intro steps so it looks more like:

```
exo model register gpt-5.4 --secret openai # I'm open to  using a hardcoded if you feel strongly though
exo repl # by default this can use a bunch of defaults
```

As a sidenote, I think the current `chat` subcommand is quite clunky, and I like the idea of having `repl` as a top-level command. Maybe eventually it could be

```
exo repl --agent X --conversation Y
```

crates/exoharness/src/basic.rs
Alexsun1one: Makes sense, thanks — agreed on keeping exoharness no-frills. Pushed an update:

- Dropped the env auto-fetch. `exo repl` now just uses whatever model you've registered (and errors with the setup steps if there isn't one). README intro matches your snippet.
- Pulled out the lazy-sandbox + head-reconcile stuff — the default repl conversation just has no shell tool now, so nothing gets provisioned. exoharness is back to untouched.
- Kept repl as a top-level command with `--agent`/`--conversation` like you suggested.

On the model — I didn't bake in a default; `exo repl` just uses whatever's registered (or `--model`), so the CLI isn't pinned to a specific model name. Felt cleaner than hardcoding one, but happy to hear if you'd rather go the other way.

Alexsun1one: One thing I noticed looking back over this: the other cli tests are all small pure/sync unit tests, but the one I added in `repl_tests.rs` spins up a real `BasicHarness` to check model resolution — so it's more of an integration-style test than its neighbors. Happy to pull the pick-a-model logic into a small pure function and test it sync so it matches the rest, or keep the integration coverage as-is if you'd rather. Whichever you prefer — lmk and I'll push it.

crates/cli/src/main.rs
```diff
@@ -312,6 +327,74 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            conversation,
+        } => {
+            let agent_slug = agent.unwrap_or_else(|| DEFAULT_REPL_SLUG.to_string());
+            let conversation_slug = conversation.unwrap_or_else(|| DEFAULT_REPL_SLUG.to_string());
+
+            let existing_agent = harness.get_agent(&agent_slug).await?;
+            if existing_agent.is_some() && model.is_some() {
+                eprintln!(
+                    "note: agent '{agent_slug}' already exists; --model only applies when creating it"
+                );
```
ankrgyl: we should probably add the log and/or tracing crates and use them. I think for now if you use https://docs.rs/log/latest/log/ it's fine, and we can switch it over to tracing at some point.

I think you can also choose not to print this message if the existing agent's model == the new model

crates/cli/src/main.rs
```diff
@@ -312,6 +327,74 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            }
+            let agent = match existing_agent {
+                Some(agent) => agent,
+                None => {
+                    let model = ensure_repl_model(harness.as_ref(), model).await?;
+                    let agent = harness
+                        .create_agent(CreateAgentRequest {
+                            slug: agent_slug,
+                            name: Some("REPL".to_string()),
```
ankrgyl: prefer lowercase.

also is the intent that each time you run `exo repl` you come back to the same session, or a new one? Most CLIs have the latter behavior but this would cause the former.

If you intend to create a new one each time, we have a pre-existing function for generating unique conversation names.

crates/cli/src/main.rs
```diff
@@ -312,6 +327,74 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                            max_output_tokens: None,
+                            max_tool_round_trips: None,
+                            braintrust: None,
+                        })
+                        .await?;
+                    println!(
+                        "created agent {} ({})",
+                        agent.record().slug,
+                        agent.record().id
+                    );
```
ankrgyl: we can probably omit this

Alexsun1one: Thanks — all good points. Pushed an update:

Sessions: exo repl now starts a fresh conversation each run (using the unique-name helper), with --conversation <slug> to resume a specific one — matches the usual CLI behavior you described.
Lowercase: done.
Dropped the "created agent" line and that --model note.
On log / tracing: agree. I left it out of this PR to avoid colliding with the logging you're adding in #4 — happy to adopt it here once that lands (and the note I'd have logged is gone now anyway).

crates/cli/src/main.rs
```diff
@@ -312,6 +327,64 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                    let mut config = conversation.config().await?;
+                    if config.shell_program.is_some() {
+                        config.shell_program = None;
+                        conversation.put_config(config).await?;
+                    }
+                    println!(
+                        "created conversation {} ({})",
+                        conversation.record().slug,
+                        conversation.record().id
+                    );
```
ankrgyl: i think we can remove this

crates/cli/src/main.rs
```diff
@@ -1207,6 +1281,46 @@ async fn list_model_bindings(
...
+        }
+        return Err(format!(
+            "model is not registered: {requested}; register it with `exo model register {requested} --secret <secret>`"
+        )
+        .into());
+    }
+    registered.first().cloned().ok_or_else(|| {
+        "no model is registered; set one up first:\n  \
+         exo secret set openai --env OPENAI_API_KEY\n  \
+         exo model register gpt-5.4 --secret openai"
```
ankrgyl: let's update to gpt-5.5

ankrgyl (commented):
Two small nits. Looking great! Will land right after these

PR #4, opened 2026-05-23
Tool support

61cygni:
First pass at tool support. Description in docs/ tools.md

Light testing via checked in irc tool, and having the agent create a curl tool for downloading and summarizing web pages. Currently have not tested with sandboxes.


examples/typescript/tools/irc.manifest.json
```diff
@@ -0,0 +1,17 @@
+{
```
ankrgyl: i'd prefer to avoid having a manifest.json and make this part of the data exposed by the object exported from typescript. we can include a small type def library that helps you define the right export shape

crates/cli/src/main.rs
```diff
@@ -177,6 +177,10 @@ enum AgentCommands {
         slug: Option<String>,
         #[arg(long)]
         module: Option<PathBuf>,
+        #[arg(long = "tool-manifest")]
+        tool_manifests: Vec<PathBuf>,
```
ankrgyl: if this is a top-level arg, then it should be a top-level feature of every harness, and therefore also be implemented in Rust.

crates/cli/src/main.rs
```diff
@@ -177,6 +177,10 @@ enum AgentCommands {
...
+        #[arg(long)]
+        disable_agent_tool_creation: bool,
```
ankrgyl: this should have a shape consistent with networking (i.e. be called "tool_creation")

crates/cli/src/main.rs
```diff
@@ -1303,6 +1367,75 @@ fn resolve_typescript_module_path(
     })
 }
 
+#[derive(serde::Deserialize)]
+struct CliToolManifest {
+    tools: Vec<CliToolManifestEntry>,
+}
+
+#[derive(serde::Deserialize)]
+struct CliToolManifestEntry {
```
ankrgyl: i think eventually we should support tools in multiple languages and the manifest extraction can be language specific

PR #7, opened 2026-05-23
CI: run rust-tests on macOS and add backend-matrix integration tests

akrentsel:
Stacked on #9 (`selectable-backends`).

## Summary

Two logical changes land:

1. **macOS added to the `rust-tests` matrix** — the existing workspace unit tests now compile-check + run on both Ubuntu and macOS.

2. **Integration-test matrix** — the (formerly Linux-only Docker) integration test is now backend-parameterised via `EXO_TEST_SANDBOX_BACKEND`, always uses the `file` secret backend with the master key inside a per-test tempdir via `XDG_CONFIG_HOME`. New workflow matrix:

   | OS    | runner            | sandbox-backend  | setup                              |
   |-------|-------------------|------------------|------------------------------------|
   | linux | ubuntu-latest     | local-process    | none                               |
   | linux | ubuntu-latest     | docker           | none (preinstalled)                |
   | macos | macos-15          | local-process    | none                               |
   | macos | macos-15-intel    | docker           | `crazy-max/ghaction-setup-docker`  |
   | ~~macos~~ | ~~macos-15~~  | ~~apple-container~~ | ~~not viable on GH-hosted runners~~ |

## Findings worth knowing

A few load-bearing decisions from iterating on the macOS arms:

- **macOS / Docker requires the Intel runner.** GitHub-hosted Apple Silicon macOS runners don't expose nested virtualization, so any docker-via-VM setup (Colima/Lima/Docker Desktop) fails at VM startup. `macos-15-intel` exposes HVF and runs Colima via `crazy-max/ghaction-setup-docker` (SHA-pinned to v5.1.0).

- **macOS / apple-container is not testable on GH-hosted runners.** Apple's `container` CLI uses VZ.framework, which needs hardware virtualization on Apple Silicon. Tracked upstream: https://github.com/actions/runner-images/issues/13505. Re-enabling requires either a self-hosted Apple Silicon runner on macOS 26+ or a third-party Mac CI service (Cirrus, MacStadium). The matrix entry is left commented out with the explanation; the install plumbing is kept intact so re-enabling is one line.

- **`container system start` prompts interactively** on first run asking whether to install the default kernel. `--enable-kernel-install` (visible in `--help`) bypasses it cleanly.

- **Unauthenticated `curl` to `api.github.com` from macOS runners hits rate limiting silently** (shared IPs, 60/h). Use `gh release download` instead, which auths via the runner's `GITHUB_TOKEN`.

- **Integration step uses `cargo test --package exo --test integration_chat`** rather than `cargo test --workspace --all-targets`. Unit tests are already covered by the `rust-tests` job in `ci.yml`; no need to recompile them here.

## Test plan

- [x] `cargo test --workspace` (51 unit tests, on Linux)
- [x] `EXO_TEST_SANDBOX_BACKEND=docker cargo test ... -- --ignored` passes locally
- [x] `EXO_TEST_SANDBOX_BACKEND=local-process cargo test ... -- --ignored` passes locally
- [x] CI: `rust-tests` on ubuntu-latest + macos-latest both green
- [x] CI: 4-cell integration matrix all green on this branch
- [x] Temporary push trigger removed before review
- [ ] Open a follow-up issue tracking macOS / apple-container CI via self-hosted runner (post-merge)

### DIFF
--- a/crates/cli/tests/integration_chat.rs
+++ b/crates/cli/tests/integration_chat.rs
@@ -1,15 +1,12 @@
 //! Integration test exercising the real `exo` binary against:
-//!   - a real Docker container (sandbox), and
+//!   - a real sandbox backend (docker / apple-container / local-process), and
 //!   - a wiremock-backed fake OpenAI Responses endpoint.
 //!
 //! `#[ignore]`'d so `cargo test` skips it by default; the integration workflow
-//! runs `cargo test --workspace -- --ignored` on push to main.
-//!
-//! Linux-only because the production sandbox backend on macOS is the Apple
-//! `container` CLI, which is not assumed to be present on macOS CI runners
-//! today. Verifying the Apple arm in CI is a separate follow-up.
-
-#![cfg(target_os = "linux")]
+//! runs `cargo test --workspace -- --ignored` and selects the backend via the
+//! `EXO_TEST_SANDBOX_BACKEND` env var (defaults to `docker`). The secret
+//! backend is always `file`, with the master key materialised inside a
+//! per-test tempdir via `XDG_CONFIG_HOME`.
 
 use std::path::PathBuf;
 use std::process::Command;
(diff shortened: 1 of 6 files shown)


akrentsel:
See successful completion here: https://github.com/ankrgyl/exo/actions/runs/26346929386

crates/cli/tests/integration_chat.rs
```diff
@@ -0,0 +1,279 @@
+//! Integration test exercising the real `exo` binary against:
```
ankrgyl, line 1: doesn't this have to be refernced in a `mod` statement somewhere to run?

akrentsel: I don't think so? The logs from the integration test run show that it gets discovered and run: https://github.com/ankrgyl/exo/actions/runs/26346929386/job/77558232782

(in the "Run integration tests" seciton, for all 4 of them...
```
running 1 test
test chat_send_round_trips_through_real_sandbox_and_mocked_openai ... ok
```
)

akrentsel: yeah, per https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery, Rust can detect the tests automatically based on directory

ankrgyl: oh neat

ankrgyl (approved):

PR #12, opened 2026-05-24
Sandbox full-state (CRIU) snapshots — blocked by docker bugs

akrentsel:
> ⚠️ **Draft. Not recommended for merge as-is.** Captures work; restore is blocked by known docker bugs across 28.x and 29.x. Keeping the branch so the code and design are reviewable, but probably either lands behind a feature flag with documented broken-restore or doesn't land at all until upstream docker fixes its checkpoint feature.

Stacked on #20 (`sandbox-snapshots-filesystem`).

## Summary

Adds a second snapshot mode on top of the filesystem-only path in #20:

- `SnapshotMode::FullState` — captures **filesystem + running processes + memory + open FDs** via `docker checkpoint create` (CRIU under the hood).
- `SnapshotKind::DockerCheckpointTar` — the tagged on-disk format alongside `DockerImageTar`.
- `/checkpoint` slash command in the REPL alongside `/snapshot`.
- `/rewind` dispatches on the persisted manifest kind, so it transparently handles either mode.

Three commits:

| # | Commit | What |
|---|---|---|
| 1 | `exoharness: full-state (CRIU) snapshot mode for docker sandboxes` | Trait extensions (`snapshot(mode)`), Docker capture+restore pipelines, `/checkpoint` REPL command, requirements preflight (`docker info ExperimentalBuild`) with an actionable error. |
| 2 | `docs: requirements doc + cover full-state snapshot mode` | New `docs/requirements.md` (general home for runtime requirements per backend) + design doc updates covering both modes and the mode/kind separation. |
| 3 | `exoharness: rewrite CRIU restore to use docker's default checkpoint dir` | Initial implementation tried `--checkpoint-dir`; docker rejects custom dirs on `start --checkpoint` ("custom checkpointdir is not supported"). Switch to docker's default `/var/lib/docker/containers/<id>/checkpoints/`, which requires `sudo tar`/`sudo mkdir` since the dir is root-owned. |

## What's verified

- Capture works end-to-end on docker 29.x: `docker checkpoint create` succeeds, the CRIU dump bytes are tarred and persisted to exoharness storage, the manifest is written correctly.
- Preflight: when `docker info --format '{{.ExperimentalBuild}}'` is `false`, `/checkpoint` returns the actionable error pointing at `docs/requirements.md` instead of letting docker fail cryptically.
- 51 unit tests pass.

## What's NOT working — known docker bugs

The restore path correctly invokes docker (verified via raw-docker reproduction without any exo code involved). Docker itself then fails:

| Path | Error | Tracking |
|---|---|---|
| Restore into fresh container, default networking | `Error response from daemon: bind-mount /proc/0/ns/net -> /var/run/docker/netns/...: no such file or directory` | Acknowledged docker bug — daemon writes invalid netns paths into the checkpoint. [Docker forum maintainer recommends `--network=host` as workaround.](https://forums.docker.com/t/using-criu-for-checkpointing-docker-containers/149448) |
| Restore in-place, any networking | `failed to upload checkpoint to containerd: commit failed: content sha256:... already exists` | [moby/moby#42900](https://github.com/moby/moby/issues/42900) — claimed fixed in 2021 (PR 47456) but regressed somewhere in docker 29.x |

Both reproduce on docker 29.4.2 (dev box) and the second one on docker 29.1.3 (test VM with full kernel modules). The forum thread reports docker 28.3.3 + CRIU + `--network=host` does work; we couldn't validate that combination because neither of the boxes we have access to runs that exact combo (one is on 29.x, the other has a stripped kernel without the modules CRIU needs).

## Why keep the code around (maybe)

- Capture is sound and the trait surface is right; when docker fixes its restore bugs or we add a different runtime (podman has substantially better CRIU integration, runc could be driven directly), the restore arm is a small addition.
- The mode/kind separation lets us cleanly add more kinds later (Apple VZ pause+save when `container` CLI exposes it, etc.).
- Documents the failure modes so future-us doesn't redo this investigation.

## Why drop the code (maybe)

- It's a meaningful diff for a feature that doesn't actually work.
- "It works on docker 28.3.3 with the right network setup" is a fragile claim we can't verify on our infrastructure.
- Filesystem snapshots cover ~99% of practical agent-rewind use cases (state lives on disk, processes are short-lived shell calls).

## Test plan

- [x] `cargo test --workspace` (51 pass)
- [x] `/checkpoint` capture succeeds; payload bytes written; manifest correct
- [x] Preflight error path verified when experimental is off
- [x] Raw-docker reproduction confirms restore failure is upstream, not in exo code
- [ ] End-to-end `/checkpoint` → `/rewind` on docker 28.3.3 with `--network=host` (not verified — no matching host available)

### DIFF
--- a/crates/cli/src/tui.rs
+++ b/crates/cli/src/tui.rs
@@ -7,7 +7,8 @@ use std::time::Duration;
 
 use executor::{
     ConversationHandle, EventData, EventId, EventQuery, EventQueryDirection, ExecutionStreamEvent,
-    HarnessConversation, SandboxId, SendRequest, SessionId, SnapshotId, StartSandboxRequest,
+    HarnessConversation, SandboxId, SendRequest, SessionId, SnapshotId, SnapshotMode,
+    StartSandboxRequest,
 };
 use lingua::universal::{UserContent, UserContentPart};
 use lingua::{Message, UniversalStreamChunk};
@@ -387,9 +388,22 @@ impl ChatRepl {
                         continue;
                     }
                     if trimmed == "/snapshot" {
-                        match self.snapshot_current_sandbox().await {
+                        match self
+                            .snapshot_current_sandbox(SnapshotMode::Filesystem)
+                            .await
+                        {
                             Ok(snapshot_id) => println!("snapshot {snapshot_id}"),
                             Err(error) => println!("snapshot failed: {error:#}"),
+                        }
+                        continue;
+                    }
+                    if trimmed == "/checkpoint" {
+                        match self
+                            .snapshot_current_sandbox(SnapshotMode::FullState)
+                            .await
+                        {
+                            Ok(snapshot_id) => println!("checkpoint {snapshot_id}"),
+                            Err(error) => println!("checkpoint failed: {error:#}"),
                         }
                         continue;
                     }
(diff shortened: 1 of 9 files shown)


akrentsel:
Woo, looks like this is working!!

image

successful snapshotting

akrentsel:
closing this for now, this is stale now.

Issue #15, opened 2026-05-24
Track and surface LLM cost/usage on message history


akrentsel:
## Motivation

The exoharness should let an agent (and its operator) see, at any point in a conversation, how much money and tokens it has spent. This unlocks budget-bounded agents — e.g. "accomplish this task in no more than \$20" — by letting the agent read its own cumulative spend out of message history and decide when to stop, compact, or escalate. It also gives operators a single source of truth for usage without re-deriving it from provider invoices.

## Current state

The data model is already partially in place but unused:

- `UsageRecord` exists in `crates/exoharness/src/types.rs:187` with fields for `prompt_tokens`, `completion_tokens`, `prompt_cached_tokens`, `prompt_cache_creation_tokens`, `completion_reasoning_tokens`, `cost_usd`, `ttft_ms`, `duration_ms`, `server_duration_ms`.
- `EventData::Messages` already carries `usage: Option<UsageRecord>` (`types.rs:237`).
- A price table and `compute_cost_usd` live in `crates/exoharness/src/pricing.rs`, covering current Anthropic and OpenAI models.

What's missing:

- **No propagation.** Every site that emits a `Messages` event passes `usage: None` (e.g. `crates/exoharness/src/basic.rs:624`, the executor's `harness_runtime.rs:402` discards usage rather than forwarding it onto the event, and all RLM/basic-executor paths). `lingua::UniversalUsage` is captured on `ModelResponse` but never converted to `UsageRecord`.
- **No aggregation surface.** No way to ask the harness "total spend / tokens for this conversation, session, or agent." Callers would have to walk events themselves.
- **No agent-facing read path.** A running executor can't introspect its own cost without re-reading and summing events by hand.

## Scope

1. **Populate `UsageRecord` at the LLM boundary.** On each successful LLM call, convert `lingua::UniversalUsage` → `UsageRecord`, fill `ttft_ms` / `duration_ms` from the existing `ModelResponse` fields, compute `cost_usd` via `pricing::compute_cost_usd`, and stamp `model` from the provider's echoed model id (falling back to the requested model). Attach it to the `Messages` event that records that turn's output.
   - Touch points: basic executor, RLM, harness runtime, and any other site that constructs `EventData::Messages` from a model response. Leave `usage: None` for synthetic / non-LLM `Messages` events (e.g. user input).
[...]

PR #16, opened 2026-05-24
history: per-message cost, tokens, and latency tracking

akrentsel:
## Summary

Adds an optional `UsageRecord` to every `EventData::Messages` event so we have a durable, per-message record of:

- **Model id** (echoed by the provider, or the requested binding as fallback)
- **Raw token counts** — prompt, completion, cached, cache-creation, reasoning
- **USD cost** — computed at call time from the LiteLLM pricing database (loaded at runtime, see below)
- **TTFT** + **wall-clock duration** — measured in the executor on both streaming and non-streaming paths
- **`server_duration_ms`** — reserved (lingua does not yet surface a provider-reported processing time)

## Pricing: why runtime LiteLLM JSON, not a hardcoded table

OpenAI and Anthropic standard APIs return tokens but not USD in their per-call responses. (OpenAI never; Anthropic only in a separate aggregate Admin API.) So cost must always be computed downstream.

Initial version hardcoded a small price table in Rust. That was a mistake — by the time I wrote the first commit, my table already had Claude Opus 4.7 at \$15/\$75 per MTok when the real price had dropped to \$5/\$25. Hand-maintained tables drift.

This version loads [LiteLLM's pricing database](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json) (2,739 model entries, community-maintained, covers all major providers + Bedrock/Azure/Vertex regional variants):

- First use: HTTP fetch into \`\$XDG_CACHE_HOME/exo/litellm_prices.json\`
- Subsequent uses (within 24h): cached
- After 24h: re-fetch, fall back to stale cache if network fails
- All fails: empty table → tokens persist, \`cost_usd: None\` (no crash)
- Env var overrides: \`EXO_LITELLM_PRICES_PATH\` (local file) and \`EXO_LITELLM_PRICES_URL\` (alternate source)

## Cached-vs-fresh: per-provider accounting matters

Different providers report cached tokens with different conventions, and getting this wrong distorts cost by up to ~10× on cache-heavy requests:

- **Anthropic-family** (`anthropic`, `bedrock_converse`, `vertex_ai-anthropic_models`, `azure_ai`): `prompt_tokens` is *fresh* input only. `cache_read` and `cache_creation` are separate. Bill all three additively.
- **OpenAI-family** (`openai`, `mistral`, etc.): `prompt_tokens` is *total* (including cached). Cached is a subset. Must subtract cached from prompt before billing fresh-input rate.

The first commit got OpenAI wrong (used the additive formula universally). This version branches on LiteLLM's \`litellm_provider\` field and applies the correct formula. Both formulas have dedicated unit tests with realistic token mixes.

## Architecture

- **\`exoharness::pricing\`** (pure data + math, no network, stays wasm-compatible):
  - \`PricingTable::from_json_str\` parses LiteLLM's schema (silently ignores entries that don't have per-token rates, like image/embedding entries).
  - \`PricingTable::lookup\` does exact + longest-prefix match (dated revisions like \`claude-sonnet-4-6-20251022\` resolve to \`claude-sonnet-4-6\`).
  - \`PricingTable::compute_cost_usd\` does per-provider math.

- **\`executor::pricing_loader\`** (network layer, gated by \`tokio::sync::OnceCell\`):
  - \`get_pricing_table()\` returns \`Arc<PricingTable>\`; loads once per process, then zero-cost.
  - Resolution: \`EXO_LITELLM_PRICES_PATH\` → cache (fresh) → fetch → stale cache → empty.

- **\`BasicExecutor::with_pricing\` + \`BasicHarness::with_pricing_table\`**: explicit-table constructors. Bypass the loader, useful for tests/embedders/air-gapped deployments.

## What's in the event JSON now

\`\`\`json
{
  \"type\": \"messages\",
  \"messages\": [...],
  \"response_id\": \"01J...\",
  \"usage\": {
    \"model\": \"claude-sonnet-4-6\",
    \"prompt_tokens\": 2847,
    \"completion_tokens\": 412,
    \"prompt_cached_tokens\": 12500,
    \"cost_usd\": 0.0146985,
    \"ttft_ms\": 842,
    \"duration_ms\": 3210
  }
}
\`\`\`

All \`usage\` sub-fields are \`Option\` + \`skip_serializing_if\`. Legacy events with no \`usage\` key continue to deserialize.

## Test plan

- [x] \`cargo test --workspace\` — 66 tests pass (was 60 on main; +11 pricing units, +1 end-to-end cost assertion, +2 backward-compat for UsageRecord JSON, -7 from refactoring the previous hardcoded-table tests away).
- [x] Pricing unit tests cover: Anthropic additive (no cache, cache hits, cache creation), OpenAI inclusive (with and without cache, subtraction semantics asserted), Bedrock regional surcharge, longest-prefix lookup, sample_spec doc entry skipped, unknown models, provider-style classification.
- [x] \`pricing_loader::tests::local_path_override_is_honored\` exercises the env-var override path.
- [x] End-to-end test \`usage_record_is_persisted_with_computed_cost\` uses \`with_pricing_table\` to inject an inline fixture — assertion is hermetic, no network dependency in CI.
- [x] \`cargo test --package exo --test integration_chat -- --ignored\` (mocked OpenAI + real local-process sandbox): unchanged, passes.
- [x] Loader exercised live: fetch succeeded, cache populated, subsequent calls cache-hit.

## Not in scope (intentional)

- **\`/cost\` REPL command** — persisting silently for now. UI surface is an easy follow-up.
- **Server-reported duration** — lingua doesn't expose this today; field reserved for when it does.
- **Cache tier resolution** — \`UniversalUsage\` collapses Anthropic's 5-min and 1-hour cache writes into one count. Cost defaults to the 5-minute rate. \`cache_creation_input_token_cost_above_1hr\` is parsed but unused.
- **Anthropic's aggregate Usage/Cost Admin API** — separate org-level endpoint, aggregate-only, can't be tied to a specific message. Not useful for per-message tracking.


akrentsel:
please don't review the code yet – I'm looking into a better way to get pricing details...

akrentsel:
This is addressing #15

akrentsel:
@ankrgyl this is ready to review, added tests

akrentsel:
@ankrgyl  this is ready to review, added tests

crates/exoharness/src/pricing.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// back to the longest entry-key that is a prefix of `model`, so
+    /// dated revisions like `claude-sonnet-4-6-20251022` resolve to
+    /// `claude-sonnet-4-6`.
+    pub fn lookup(&self, model: &str) -> Option<&ModelEntry> {
+        if let Some(entry) = self.entries.get(model) {
+            return Some(entry);
+        }
+        self.entries
+            .iter()
+            .filter(|(key, _)| model.starts_with(key.as_str()))
```
Alexsun1one: prefix-match without a separator boundary can fall through in surprising ways. e.g. `gpt-4o-mini` matches `gpt-4o` (intended) but also matches `gpt-4` (not intended), since `"gpt-4o".starts_with("gpt-4")` is true; if the more specific entry is missing the longest-prefix winner is still wrong.

if the goal is the `claude-sonnet-4-6-20251022 -> claude-sonnet-4-6` case, tightening the filter to require `model[key.len()..]` be empty or start with `-` / `:` keeps that behavior and stops `gpt-4o` from sliding into `gpt-4` when an entry is absent.

Alexsun1one (commented):
two small things on the price lookup side, otherwise this looks great. the LiteLLM-as-source-of-truth + per-provider accounting story is the right call, and i like that `with_pricing` keeps embedders/air-gapped runs honest. inline notes below.

crates/exoharness/src/pricing.rs
```diff
@@ -0,0 +1,426 @@
...
+
+impl ProviderStyle {
+    fn for_litellm_provider(provider: Option<&str>) -> Self {
+        // LiteLLM uses these strings (see model_prices_and_context_window.json
+        // values for litellm_provider). Anything not classified here defaults
+        // to Inclusive — same convention as OpenAI, which is the safer of the
+        // two when cached==0 (the typical case).
+        match provider {
+            Some(p) if p.starts_with("anthropic") => Self::Additive,
+            Some(p) if p.starts_with("bedrock") => Self::Additive,
```
Alexsun1one: this catches `bedrock_converse` correctly but also catches plain `bedrock` provider entries, which on Bedrock covers Mistral, Cohere, Meta Llama, AI21, and friends. those follow OpenAI-style inclusive `prompt_tokens`, not Anthropic-style additive.

the real-world impact is small today because cache_read on non-Anthropic Bedrock models is uncommon. but if/when those providers add caching, the formula will over-count fresh-input tokens.

probably `starts_with("bedrock_converse")` only, or an explicit allow-list of bedrock-anthropic provider strings.

crates/executor/src/pricing_loader.rs
```diff
@@ -0,0 +1,237 @@
...
+                }
+            }
+        })
+        .await
+        .clone()
+}
+
+async fn try_load() -> anyhow::Result<PricingTable> {
+    // 1. Local path override — used by tests and air-gapped setups.
+    if let Ok(path) = std::env::var("EXO_LITELLM_PRICES_PATH") {
```
ankrgyl: i prefer all env vars to be parsed through clap, so that someone could propagate the pricing table as a CLI arg too. it also forces all functions (like this one) to be relatively pure

maybe we add this to a skill in the repo somewhere?

ankrgyl (approved):

akrentsel:
rewriting in #54

PR #20, opened 2026-05-25
Sandbox filesystem snapshots and rewind (Docker)

akrentsel:
Fills in the snapshot/rewind plumbing that was previously metadata-only. **Captures and restores a sandboxed container's filesystem state.** A separate PR (#12) stacks on top of this one for an experimental CRIU full-state path — that one is intentionally a draft because it's blocked by upstream docker bugs.

## What this does

Inside a chat REPL (`exo chat repl <agent> <conv>`), three new slash commands:

| Command | What it does |
|---|---|
| `/snapshot` | Capture the sandbox's **filesystem state** as it is right now — every file created, modified, or deleted since the base image. Prints a snapshot UUID. On disk this is a `docker save` tarball under `conversations/<conv>/snapshots/<id>/payload.bin` plus a `manifest.json` sidecar. |
| `/snapshots` | List every snapshot taken in this conversation. |
| `/rewind <uuid>` | Stop the current container, bring up a fresh one whose filesystem matches the snapshot. Subsequent tool calls see the rolled-back state. |

## What this does NOT do — be clear about the limit

The snapshot is **filesystem only**. Specifically:

- ✅ **Files persisted to disk inside the container** are captured and restored — files you created, packages you installed, configs you wrote.
- ❌ **Running processes are not preserved.** If you `/snapshot` while `nohup sleep 9999 &` is running and then `/rewind`, the file `/proc/<pid>` is not coming back; the process is gone. The new container boots fresh.
- ❌ **In-memory state is gone.** An interactive REPL's variables, an open TCP connection, a buffered write that hadn't been flushed — none of these survive a rewind.
- ❌ **Conversation history is not rewound.** Your chat messages and event log stay where they were. Use `conversation fork` if you want to rewind the conversation itself; snapshots only operate on the sandbox filesystem underneath.

For agent workflows where "state worth preserving" = "files written to disk + tools/packages installed", filesystem snapshots cover the case. For pause-and-resume of long-running in-memory processes, you'd need full-state (CRIU), which is #12's beat and currently impractical to ship due to upstream docker bugs.

## Demo

gif:
snapshot-rewind

```bash
# Bootstrap a state dir + agent + conversation, drop into the REPL.
# (--networking enabled because the demo needs the LLM to reach OpenAI)
STATE=/tmp/exo-snapshot-demo && rm -rf $STATE && mkdir -p $STATE && \
EXO="./target/debug/exo --root $STATE --secret-backend file --sandbox-backend docker --master-key-path $STATE/master.key" && \
$EXO secret set OPENAI_KEY --env OPENAI_API_KEY && \
$EXO model register gpt-4o-mini --secret OPENAI_KEY && \
$EXO agent create demo --model gpt-4o-mini --networking enabled && \
$EXO conversation create demo first --repl
```

Then inside `first>`:

```
create /tmp/demo.txt with the content "version 1" and cat it back
```
*Sandbox now has `/tmp/demo.txt` with "version 1".*

```
/snapshot
```
> `snapshot 019e5782-7c6b-72a2-b4fa-a81bf56eb37e`
>
> Behind the scenes: `docker commit -p` → `docker save` → tarball (~47MB for the ubuntu:24.04 base) → written to `conversations/<conv>/snapshots/<id>/payload.bin` + `manifest.json`. The local docker image is dropped (`docker image rm`) — exoharness storage is the canonical home.

```
overwrite /tmp/demo.txt with "version 2" and cat it back
```
*Sandbox file now reads "version 2".*

```
/snapshots
```
> ```
> SNAPSHOT                              SANDBOX
> 019e5782-7c6b-72a2-b4fa-a81bf56eb37e  sandbox-019e5782-2a46-7970-a5bf-62900a2233e8
> ```

```
/rewind 019e5782-7c6b-72a2-b4fa-a81bf56eb37e
```
> `rewound to snapshot 019e5782-7c6b-72a2-b4fa-a81bf56eb37e`
>
> Behind the scenes: `docker load < payload.bin` → fresh container booted from the restored image → swapped into the warm pool, keyed identically. Mounts / network policy / lifecycle preserved from the original sandbox request.

```
cat /tmp/demo.txt
```
> `version 1`   ← the rewind worked

You can take many snapshots in a conversation and rewind to any of them.

## Six commits

| # | Commit | What |
|---|---|---|
| 1 | `exoharness: sandbox snapshot/restore trait surface and Docker implementation` | Adds `SnapshotPayload { kind, bytes }` + `SnapshotKind::DockerImageTar`. Extends `ManagedSandboxHandle::snapshot()` and `ManagedSandboxBackend::acquire_from_snapshot(req, payload)`. Stubs with explicit "not supported" errors for OneShot / LocalProcess / AppleContainer with clear "where the real impl goes" comments. |
| 2 | `exoharness: persist sandbox snapshots and restore via start_sandbox` | Wires `snapshot_sandbox` to actually capture and persist. Wires `start_sandbox` to load the payload by manifest kind and call `acquire_from_snapshot`. |
| 3 | `cli: /snapshot, /snapshots, /rewind slash commands in chat repl` | The user-facing surface. Lives in the REPL because the sandbox handle is per-process; a top-level `exo conversation snapshot` subcommand needs cross-invocation container adoption (separate follow-up). |
| 4 | `docs: sandbox snapshot/rewind design` | `docs/sandbox-snapshots.md` covering data flow, on-disk layout, backend extension story, known limits. |
| 5 | `exoharness: default sandbox image to ubuntu:24.04` | debian:bookworm doesn't ship `procps`; even basic "list running processes" tool calls hit `command not found`. Ubuntu 24.04 has procps + coreutils in the base. |
| 6 | `ci: end-to-end snapshot + rewind round-trip test (docker, linux)` | New `crates/cli/tests/snapshot_round_trip.rs` that drives the harness library directly against real Docker. Mirrors the demo's lifecycle and asserts on two independent rewind signals (file content rolls back AND a post-snapshot file disappears). Wired into the integration matrix workflow. |

## On-disk shape

```
agents/<agent_id>/conversations/<conv_id>/snapshots/<snapshot_id>/
├── manifest.json   { snapshot_id, sandbox_id, kind, created_at, payload_size_bytes }
└── payload.bin     docker save tarball for SnapshotKind::DockerImageTar
```

The snapshot's existence is also recorded in the conversation event log as `SandboxSnapshotted { sandbox_id, snapshot_id }`, which is what `/snapshots` walks to render the listing.

## Adding more backends

Anyone adding snapshot support for a new sandbox backend follows this recipe:

1. Add a new `SnapshotKind` variant naming the on-disk format (e.g. `AppleContainerImageTar`).
2. Implement `ManagedSandboxHandle::snapshot` to produce that kind. The Docker version is the template — three CLI calls and a `Bytes` capture.
3. Implement `ManagedSandboxBackend::acquire_from_snapshot` to consume the same kind, with an explicit kind-mismatch error.
4. Backends that genuinely can't snapshot (local-process today) keep returning an explicit error.

No other layer changes. The conversation orchestration, on-disk layout, and CLI surface are all backend-agnostic.

## Test plan

- [x] `cargo test --workspace` (51 unit tests pass)
- [x] **End-to-end CI test added: `crates/cli/tests/snapshot_round_trip.rs`** — drives the harness library directly against real Docker on Linux, asserts the file content rolls back AND a post-snapshot file disappears after `/rewind`. Passes locally in ~1.8s; self-skips on non-docker matrix cells. Wired into the existing `integration.yml` workflow alongside the existing `integration_chat` test.
- [x] Manual REPL verification (the demo above): `/snapshot` → file content modified → `/rewind` → file content rolled back. Verified visually before the automated test was written.
- [x] On-disk layout (manifest.json + payload.bin) verified by the round-trip test using `read_dir()`.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -112,7 +112,13 @@ jobs:
           container system status
 
       - name: Run integration tests
-        # Narrow to the one integration target — unit tests are already
-        # covered by the rust-tests job in ci.yml, no need to recompile or
-        # rerun them here.
-        run: cargo test --package exo --test integration_chat -- --ignored
+        # List each integration-test target explicitly. Each scenario lives
+        # in its own file under crates/cli/tests/ and self-skips when its
+        # required backend isn't available on the current matrix cell.
+        # Adding a new scenario in a future PR means appending another
+        # `--test <name>` flag here.
+        run: |
+          cargo test --package exo \
+            --test integration_chat \
+            --test snapshot_round_trip \
+            -- --ignored
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -112,13 +112,6 @@ jobs:
           container system status
 
       - name: Run integration tests
-        # List each integration-test target explicitly. Each scenario lives
-        # in its own file under crates/cli/tests/ and self-skips when its
-        # required backend isn't available on the current matrix cell.
-        # Adding a new scenario in a future PR means appending another
-        # `--test <name>` flag here.
-        run: |
-          cargo test --package exo \
-            --test integration_chat \
-            --test snapshot_round_trip \
-            -- --ignored
+        # `--tests` auto-picks-up new test files; each self-skips on
+        # cells that don't have its required backend.
+        run: cargo test --package exo --tests -- --ignored
--- a/crates/cli/src/tui.rs
+++ b/crates/cli/src/tui.rs
@@ -7,7 +7,7 @@ use std::time::Duration;
 use anyhow::Result;
 use executor::{
     ConversationHandle, EventData, EventId, EventQuery, EventQueryDirection, ExecutionStreamEvent,
-    HarnessConversation, SendRequest, SessionId,
+    HarnessConversation, SandboxId, SendRequest, SessionId, SnapshotId, StartSandboxRequest,
 };
 use lingua::universal::{UserContent, UserContentPart};
 use lingua::{Message, UniversalStreamChunk};
@@ -380,6 +380,44 @@ impl ChatRepl {
                         self.print_transcript().await?;
                         continue;
                     }
+                    if trimmed == "/help" {
+                        print_help();
+                        continue;
+                    }
+                    if trimmed == "/snapshot" {
+                        match self.snapshot_current_sandbox().await {
+                            Ok(snapshot_id) => println!("snapshot {snapshot_id}"),
+                            Err(error) => println!("snapshot failed: {error:#}"),
+                        }
+                        continue;
+                    }
+                    if trimmed == "/snapshots" {
+                        match self.list_snapshots().await {
+                            Ok(snapshots) if snapshots.is_empty() => {
+                                println!("no snapshots yet for this conversation");
+                            }
+                            Ok(snapshots) => {
+                                println!("SNAPSHOT\tSANDBOX");
+                                for (snapshot_id, sandbox_id) in snapshots {
+                                    println!("{snapshot_id}\t{sandbox_id}");
+                                }
+                            }
+                            Err(error) => println!("listing snapshots failed: {error:#}"),
+                        }
+                        continue;
+                    }
+                    if let Some(arg) = trimmed.strip_prefix("/rewind ") {
+                        let snapshot_id_str = arg.trim();
+                        if snapshot_id_str.is_empty() {
+                            println!("usage: /rewind <snapshot-id>");
+                            continue;
+                        }
+                        match self.rewind_to_snapshot(snapshot_id_str).await {
+                            Ok(()) => println!("rewound to snapshot {snapshot_id_str}"),
+                            Err(error) => println!("rewind failed: {error:#}"),
+                        }
+                        continue;
+                    }
 
                     self.editor.add_history_entry(line.as_str())?;
                     self.send(trimmed).await?;
@@ -400,6 +438,50 @@ impl ChatRepl {
             self.conversation.close_session(session_id).await?;
         }
 
+        Ok(())
+    }
+
+    /// Find the most recent `SandboxCreated` event and snapshot that sandbox.
+    /// The conversation's active sandbox must have been created (or started)
+    /// in *this* REPL session — running_sandboxes is a per-process map.
+    async fn snapshot_current_sandbox(&self) -> Result<SnapshotId> {
+        let sandbox_id = latest_sandbox_id(self.conversation.as_ref())
+            .await?
+            .ok_or_else(|| {
+                anyhow::anyhow!("no sandbox has been created in this conversation yet")
+            })?;
+        let id = self
+            .conversation
+            .exoharness_handle()
+            .snapshot_sandbox(sandbox_id)
+            .await?;
+        Ok(id)
+    }
+
+    async fn list_snapshots(&self) -> Result<Vec<(SnapshotId, SandboxId)>> {
+        list_snapshots(self.conversation.as_ref()).await
+    }
+
+    /// Restore the conversation's sandbox to a previously-taken snapshot.
+    /// Stops the current container, decodes the snapshot payload, and starts
+    /// a fresh container from that state.
+    async fn rewind_to_snapshot(&self, snapshot_id_str: &str) -> Result<()> {
+        let snapshot_id = snapshot_id_str
+            .parse::<SnapshotId>()
+            .map_err(|error| anyhow::anyhow!("invalid snapshot id `{snapshot_id_str}`: {error}"))?;
+        let sandbox_id = sandbox_id_for_snapshot(self.conversation.as_ref(), snapshot_id)
+            .await?
+            .ok_or_else(|| {
+                anyhow::anyhow!("snapshot {snapshot_id} not found in this conversation")
+            })?;
+        self.conversation
+            .exoharness_handle()
+            .start_sandbox(StartSandboxRequest {
+                id: sandbox_id,
+                snapshot_id,
+                idle_seconds: None,
+            })
+            .await?;
         Ok(())
     }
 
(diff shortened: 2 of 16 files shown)


crates/exoharness/src/sandbox.rs
```diff
@@ -103,14 +125,29 @@ pub trait ManagedSandboxHandle: Send + Sync {
...
+    /// the supplied `payload.kind`.
+    async fn acquire_from_snapshot(
+        &self,
+        request: SandboxRequest,
+        payload: SnapshotPayload,
+    ) -> Result<Arc<dyn ManagedSandboxHandle>>;
 }
 
-pub const DEFAULT_SANDBOX_IMAGE: &str = "docker.io/library/debian:bookworm";
+pub const DEFAULT_SANDBOX_IMAGE: &str = "docker.io/library/ubuntu:24.04";
```
akrentsel, line 150: note: I changed the default to be thicker. I think this will match what people expect more? any reason we were defaulting to such a thin option before?

docs/sandbox-snapshots.md
```diff
@@ -0,0 +1,240 @@
...
+producer/consumer interface (`AsyncRead`/`AsyncWrite`) is a clean
+follow-up if larger images become routine.
+
+### Snapshot lifecycle
+
+There is no GC. Snapshots remain on disk until the conversation directory
+is deleted. A future addition could prune snapshots older than the most
+recent N, or evict by total size.
+
+### Restore semantics
```
akrentsel, line 233: Note: this changes in the subsequent PR, #21. Leaving this here for accuracy in this PR.

crates/exoharness/src/sandbox.rs
ankrgyl: just minimalism but i think it's fine to change the default, and someone who wants a more minimal option can just use that

.github/workflows/integration.yml
```diff
@@ -112,7 +112,13 @@ jobs:
...
       - name: Run integration tests
-        # Narrow to the one integration target — unit tests are already
-        # covered by the rust-tests job in ci.yml, no need to recompile or
-        # rerun them here.
-        run: cargo test --package exo --test integration_chat -- --ignored
+        # List each integration-test target explicitly. Each scenario lives
+        # in its own file under crates/cli/tests/ and self-skips when its
+        # required backend isn't available on the current matrix cell.
+        # Adding a new scenario in a future PR means appending another
+        # `--test <name>` flag here.
```
ankrgyl: is it possible to make this auto discovered? I highly doubt people will remember to update this file

crates/cli/src/tui.rs
```diff
@@ -382,6 +382,44 @@ impl ChatRepl {
                         self.print_transcript().await?;
                         continue;
                     }
+                    if trimmed == "/help" {
```
ankrgyl: maybe we should switch to `match trimmed`?

crates/cli/src/tui.rs
```diff
@@ -382,6 +382,44 @@ impl ChatRepl {
...
+                            Err(error) => println!("listing snapshots failed: {error:#}"),
+                        }
+                        continue;
+                    }
+                    if let Some(arg) = trimmed.strip_prefix("/rewind ") {
+                        let snapshot_id_str = arg.trim();
+                        if snapshot_id_str.is_empty() {
+                            println!("usage: /rewind <snapshot-id>");
+                            continue;
+                        }
```
ankrgyl: may also want to check if it contains whitespace (i.e. someone provided multiple args) and throw an error in that case

crates/cli/src/tui.rs
```diff
@@ -561,6 +639,91 @@ fn render_user_content_for_history(content: &UserContent) -> String {
...
+/// created yet (e.g. nothing has been chatted with).
+async fn latest_sandbox_id(
+    conversation: &dyn HarnessConversation,
+) -> Result<Option<SandboxId>, Box<dyn Error>> {
+    let result = conversation
+        .exoharness_handle()
+        .get_events(Some(EventQuery {
+            cursor: None,
+            direction: Some(EventQueryDirection::Desc),
+            limit: Some(50),
```
ankrgyl: If there's a filter on the type, then do we need 50 or just 1?

crates/cli/src/tui.rs
```diff
@@ -561,6 +639,91 @@ fn render_user_content_for_history(content: &UserContent) -> String {
...
+            session_id: None,
+            turn_id: None,
+            types: Some(vec!["sandbox_created".to_string()]),
```
ankrgyl: wonder if we should make types accept an enum so we don't typo `sandbox_created`

crates/cli/src/tui.rs
```diff
@@ -561,6 +639,91 @@ fn render_user_content_for_history(content: &UserContent) -> String {
     }
 }
 
+fn print_help() {
```
ankrgyl, line 648: separate PR probably but we should likely switch to a CLI library

crates/cli/src/tui.rs
```diff
@@ -382,6 +382,44 @@ impl ChatRepl {
                         self.print_transcript().await?;
                         continue;
                     }
+                    if trimmed == "/help" {
+                        print_help();
+                        continue;
+                    }
+                    if trimmed == "/snapshot" {
```
ankrgyl: should we support `snapshot <id>` as well? Imagine you have two active sandboxes

crates/cli/src/tui.rs
```diff
@@ -561,6 +639,91 @@ fn render_user_content_for_history(content: &UserContent) -> String {
...
+                direction: Some(EventQueryDirection::Asc),
+                limit: Some(100),
+                session_id: None,
+                turn_id: None,
+                types: Some(vec!["sandbox_snapshotted".to_string()]),
+            }))
+            .await?;
+        let events_empty = result.events.is_empty();
+        for event in result.events {
+            if let EventData::SandboxSnapshotted {
```
ankrgyl: if it isn't this event type should we throw an error? Maybe `get_events` should have a convenience function called `get_events_by_type` which returns that type always

crates/exoharness/src/basic.rs
```diff
@@ -990,9 +992,43 @@ impl ConversationHandle for BasicConversationHandle {
...
+            .ok_or_else(|| anyhow!("sandbox {id} is not running; start it before snapshotting"))?;
+        let payload = handle.snapshot().await?;
+
         let _guard = self.harness.inner.write_lock.lock().await;
         let mut sandbox = self.load_sandbox(&id).await?;
         let snapshot_id = Uuid7::now();
+
+        let manifest = StoredSnapshotManifest {
+            snapshot_id,
+            sandbox_id: id.clone(),
```
ankrgyl, line 946: i think Uuid7 is `Copy` (might be wrong). Meaning you can omit the `clone()`

crates/exoharness/src/basic.rs
```diff
@@ -990,9 +992,43 @@ impl ConversationHandle for BasicConversationHandle {
...
+        self.harness
+            .inner
+            .storage
+            .put_bytes(snapshot_dir.join("payload.bin"), payload.bytes.to_vec())
+            .await?;
+        self.harness
+            .inner
+            .storage
+            .put_json(snapshot_dir.join("manifest.json"), &manifest)
+            .await?;
```
ankrgyl: these two writes should be parallelized

crates/exoharness/src/basic.rs
```diff
@@ -1023,6 +1059,35 @@ impl ConversationHandle for BasicConversationHandle {
...
+            .get_json(snapshot_dir.join("manifest.json"))
+            .await
+            .with_context(|| {
+                format!(
+                    "loading snapshot manifest for {} (have you taken a snapshot?)",
+                    request.snapshot_id
+                )
+            })?;
+        let payload_bytes = self
+            .harness
```
ankrgyl: same -- these two reads should be parallelized

crates/exoharness/src/sandbox.rs
```diff
@@ -1172,3 +1315,133 @@ fn new_warm_container_name(key: &SandboxKey) -> String {
...
+        .arg(&snap_tag)
+        .output()
+        .await;
+    if let Ok(output) = &rm_output
+        && !output.status.success()
+    {
+        eprintln!(
+            "warning: failed to remove ephemeral snapshot image {snap_tag}: {}",
+            render_command_error(&output.stderr)
+        );
```
ankrgyl, line 1390: i think we are nearing a point where we should add proper logging. i can do this in a separate change

crates/exoharness/src/sandbox.rs
```diff
@@ -1172,3 +1315,133 @@ fn new_warm_container_name(key: &SandboxKey) -> String {
...
+        bail!(
+            "docker load failed: {}",
+            render_command_error(&output.stderr)
+        );
+    }
+
+    // docker load prints lines like:
+    //   Loaded image: <ref>
+    //   Loaded image ID: sha256:<digest>
+    // Prefer the named-tag line; fall back to image-ID.
```
ankrgyl, line 1438: should we consider using https://github.com/fussybeaver/bollard or similar to deal with Docker? I also think you can pass `--json` to most commands and get a structured output

docs/sandbox-snapshots.md
```diff
@@ -0,0 +1,240 @@
+# Sandbox Snapshots
```
ankrgyl, line 1: separately, we should think about how to expose these docs on the site

crates/cli/src/tui.rs
akrentsel: Yeah, at this point we're starting to hack one together. I'll file an issue to get back to that...filed as #32

crates/cli/src/tui.rs
akrentsel: yup. done.

crates/cli/src/tui.rs
akrentsel: hmmm, yes that is a possible case. added support there.

when i sit down to write up a paper form of exoharness in the next few weeks, I want to make this mental model more concrete though – the multiple concurrent sandbox case doesn't naturally fit my mental model.

crates/exoharness/src/basic.rs
akrentsel: the SandboxId is string type! https://github.com/ankrgyl/exo/blob/023d65c22911997047a7288dd3fde4d6e127e1ed/crates/exoharness/src/types.rs#L415 so still requires clone

crates/exoharness/src/basic.rs
akrentsel: done

crates/exoharness/src/basic.rs
akrentsel: done

docs/sandbox-snapshots.md
akrentsel: Yes, agreed. I like the template that the harbor folks use for their documentation, https://www.harborframework.com/docs. (and other projects, like skyrl: https://docs.skyrl.ai/docs). 

we can have a separate docs page linked from the main exoharness.ai landing page. To be addressed later, it's probably fine for it to be in markdown files for now.

crates/exoharness/src/sandbox.rs
akrentsel: SG. filed #33 to track.

.github/workflows/integration.yml
akrentsel: Yeah thanks, that's a good point. done.

crates/cli/src/tui.rs
akrentsel: Oops yes – just the first will be correct. Fixed.

crates/cli/src/tui.rs
akrentsel: Yeah, definitely. Made that change (in this same PR, hopefully not too much additional to review)

crates/cli/src/tui.rs
akrentsel: Yes, throwing the error makes sense. Done (here and elsewhere). Mind if I leave get_events_by_type to a follow-up PR?

crates/exoharness/src/sandbox.rs
akrentsel: Cool, I didn't think to look for a rust library for this. Yes, that would be better. Let me look into the right way to do this in a standalone PR. Filed #34 to track.

akrentsel:
Comments addressed. A handful of followup tasks came up, filed in their own issues.

ankrgyl (approved):

PR #21, opened 2026-05-25
container lifecycle: resume sandboxes across exo processes

akrentsel:
## What this changes

Container lifecycle becomes resume-aware. Before this PR, every `exo` invocation created a fresh sandbox; after it, an agent continuing a conversation picks up the same sandbox it was using.

## Lifecycle on every `run_in_sandbox`

```
when the conversation has a recorded sandbox_id:

    (a) container is running
            └─► attach and use it

    (b) container exists but is stopped
            └─► `docker start` it, then use it

    (c) container is gone (rm'd, TTL-expired, server-side auto-cleanup)
            │
            ├─ (i)  conversation log has a snapshot for this sandbox
            │       └─► restore the latest snapshot, use it
            │
            └─ (ii) no snapshot ever taken
                    └─► create a fresh container from base image

otherwise (no recorded sandbox_id):
    └─► create a fresh container from base image
```

Tiers (a) and (b) are unified inside `backend.try_resume`. Tier (c)(i) goes through the existing `start_sandbox` / `acquire_from_snapshot` path that #20 introduced. Tier (c)(ii) is the current `create_sandbox` path.

## What changed

| Layer | Change |
|---|---|
| trait | new `ManagedSandboxBackend::try_resume(req) -> Option<Handle>`: "find the sandbox for this `SandboxKey`, start it if stopped, attach if running, return `None` otherwise." |
| Docker / AppleContainer | implements `try_resume` via `docker ps -a --filter label=exo.sandbox.key=…` (and `container list --format json` for Apple), matches by spec-hash, reaps drifted entries, enforces cross-process idle TTL via `docker inspect .State.FinishedAt`. |
| LocalProcess | `try_resume` always `Ok(None)` — no persistent identity to resume. |
| `Drop` | `Drop` for `CliContainerSandboxBackend` now does `docker stop -t 0` instead of `docker rm -f`. Containers survive process exit, labelled and ready for the next exo invocation to find. |
| `run_in_sandbox` | on in-memory cache miss, calls `backend.try_resume` and binds the handle. The old error `"sandbox is not active in this process"` only fires now when the cross-process lookup also misses. |
| `ensure_shell_sandbox` | runs the 3-tier chain above. Tier 2 (snapshot restore) walks `SandboxSnapshotted` events for the recorded sandbox_id. |
| tests | one new integration test: two cross-process `chat send` invocations against one conversation produce exactly one docker container. |

## User-visible behavior change

- **Before:** exit exo → all warm containers `docker rm -f`'d. `docker ps -a` was clean after exit.
- **After:** exit exo → warm containers stopped (Exited state on host). `docker ps -a` shows them. They get cleaned up the next time exo runs a sandbox that's been idle past the TTL (5 min default, the existing `idle_seconds=300` from `ensure_shell_sandbox`).

If a user previously relied on "exit exo, zero residue on the host," this is a behavior change. The trade-off is what enables resume; manual `docker rm` still works for anyone who wants the old behavior on demand.

## Stack

```
main
  └── #20  sandbox-snapshots-filesystem    Docker snapshot + rewind
        └── #21  this PR                   Cross-process resume + 3-tier fallback
              └── (follow-up)              Daytona backend, inherits the resume contract
```

## Test plan

- [x] `cargo test --workspace` — 51 unit tests pass
- [x] `EXO_TEST_SANDBOX_BACKEND=docker cargo test -- --ignored` — both integration tests pass (existing one with updated container-state assertion, and the new cross-process resume test)
- [x] Manual end-to-end: `chat send 'echo foo > /tmp/x.txt'` then in a separate exo invocation `chat send 'cat /tmp/x.txt'` — the file persists, same container is resumed.

### DIFF
--- a/crates/executor/src/harness_tool.rs
+++ b/crates/executor/src/harness_tool.rs
@@ -2,8 +2,8 @@ use crate::{AgentConfig, ConversationConfig, ToolRuntime};
 use async_trait::async_trait;
 use exoharness::{
     ConversationHandle, CreateSandboxRequest, DEFAULT_SANDBOX_IMAGE, EventData, EventKind,
-    EventQuery, EventQueryDirection, FileSystemMount, FileSystemMountMode, Result,
-    RunInSandboxRequest, SandboxProvider, ToolRequest, ToolResult,
+    EventQueryDirection, FileSystemMount, FileSystemMountMode, Result, RunInSandboxRequest,
+    SandboxProvider, SnapshotId, StartSandboxRequest, ToolRequest, ToolResult,
 };
 use futures::io::AsyncReadExt;
 use serde::{Deserialize, Serialize};
@@ -124,6 +124,8 @@ pub(crate) async fn ensure_shell_sandbox(
                 return Ok(sandbox.id);
             };
 
+            // Tier 1: container is still around — `run_in_sandbox` does the
+            // cross-process `try_resume` internally on cache miss.
             let healthcheck = conversation
                 .run_in_sandbox(RunInSandboxRequest {
                     id: sandbox.id.clone(),
@@ -134,9 +136,26 @@ pub(crate) async fn ensure_shell_sandbox(
             if healthcheck.is_ok() {
                 return Ok(sandbox.id);
             }
+
+            // Tier 2: container is gone. If we ever took a snapshot of this
+            // sandbox in this conversation, restore from the latest one.
+            if let Some(snapshot_id) =
+                latest_snapshot_for_sandbox(conversation, &sandbox.id).await?
+                && conversation
+                    .start_sandbox(StartSandboxRequest {
+                        id: sandbox.id.clone(),
+                        snapshot_id,
+                        idle_seconds: Some(sandbox.idle_seconds),
+                    })
+                    .await
+                    .is_ok()
+            {
+                return Ok(sandbox.id);
+            }
         }
     }
 
+    // Tier 3: nothing reusable; create fresh.
     conversation
         .create_sandbox(CreateSandboxRequest {
             provider: desired_provider,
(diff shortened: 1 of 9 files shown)


akrentsel:
Added tests, shown passing here: https://github.com/ankrgyl/exo/actions/runs/26390452231/job/77678665189

crates/executor/src/harness_tool.rs
```diff
@@ -151,6 +178,34 @@ pub(crate) async fn ensure_shell_sandbox(
...
 
+async fn latest_snapshot_for_sandbox(
+    conversation: &dyn ConversationHandle,
+    sandbox_id: &str,
+) -> Result<Option<SnapshotId>> {
+    let events = conversation
+        .get_events(Some(EventQuery {
+            cursor: None,
+            direction: Some(EventQueryDirection::Desc),
+            limit: Some(100),
```
ankrgyl: same questions as the other PR: do we need 100 of these? And should we make a more ergonomic helper

crates/exoharness/src/sandbox.rs
```diff
@@ -1289,6 +1398,226 @@ fn is_missing_container_error(stderr: &str) -> bool {
...
+            "docker ps failed while listing resumable containers: {}",
+            render_command_error(&output.stderr)
+        ));
+    }
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let mut matches: Vec<(String, String, String)> = Vec::new();
+    let mut stale_names: Vec<String> = Vec::new();
+    for line in stdout.lines() {
+        let mut parts = line.splitn(4, '\t');
```
ankrgyl: this seems quite brittle. Have you tried using `--format json`?

akrentsel: yeah swapped to --format json here (and elsewhere, using going forward).

crates/executor/src/harness_tool.rs
akrentsel: Ok, fixed this with a cleaner helper. We do need more than 1 of these here – because we have two types of filtering happening (in the new code). First we filter down to snapshot events, and THEN closure is applied to get the snapahot of this specific sandbox. there could be other sandbox snapshots in that same history. picked 100 as a limit, but we can pick some other lower const if you preferred.

akrentsel:
Ready for review here: @ankrgyl

PR #22, opened 2026-05-25
Daytona remote-container sandbox backend

akrentsel:
A new sandbox backend that runs containers on [Daytona](https://daytona.io/) instead of locally. Selected via `--sandbox-backend daytona` (env: `EXO_SANDBOX_BACKEND=daytona`). All the harness machinery — tool dispatch, snapshot/rewind, the cross-process resume contract from #21 — works against it transparently.

```bash
export DAYTONA_API_KEY=...
export DAYTONA_ORGANIZATION_ID=...
exo --sandbox-backend daytona chat repl my-agent my-convo
```

## Inheriting the resume contract

The lifecycle work in #21 was designed so Daytona slots in cleanly:

- `ManagedSandboxBackend::acquire` → `POST /api/sandbox` (create fresh).
- `ManagedSandboxBackend::try_resume` → label-based lookup against Daytona's control plane, `POST /api/sandbox/{id}/start` if stopped. Daytona's server-side `stop` preserves filesystem state, so resume comes essentially free.
- `ManagedSandboxHandle::stop` → `POST /api/sandbox/{id}/stop` (not delete).
- `ManagedSandboxHandle::snapshot` → `POST /api/sandbox/{id}/snapshot`. Returns a `SnapshotKind::DaytonaSnapshot` payload whose bytes are a small JSON manifest pointing at the named Daytona snapshot (**bytes-by-reference**, in contrast to Docker's `DockerImageTar` bytes-by-value). No multi-MB tarball ever crosses the wire.

End-to-end verified live: two `chat send` invocations against the same conversation, in separate exo processes, write a marker file in send 1, read it back in send 2 — same Daytona sandbox UUID, file persists.

## Snapshot status: code is in, endpoint is not

Discovered while live-testing: the snapshot path is **upstream-unavailable** at the moment, not a code defect.

The code follows Daytona's published OpenAPI spec exactly:
- `POST /api/sandbox/{sandboxIdOrName}/snapshot`, body `{"name": "..."}` — matches `CreateSandboxSnapshot` in their `api-json`.
- The flag is consistent with the SDK naming: TypeScript SDK calls it `_experimental_createSnapshot` (note the leading underscore).

But the endpoint returns `404 Cannot POST` on `app.daytona.io`. Cross-checked three ways, all hit the same wall:
1. Raw curl with the spec-matching body shape → 404.
2. exo `/snapshot` slash command → same 404 surfaced through to the user.
3. **Daytona's official Python SDK** calling `sandbox._experimental_create_snapshot(...)` → `DaytonaNotFoundError: Failed to create snapshot: Cannot POST /api/sandbox/<id>/snapshot`.

So the operation is documented + present in SDKs but not actually deployed on the public API host yet. When Daytona enables it our code will Just Work without modification.

Related harness gap noticed while testing: `ConversationHandle::snapshot_sandbox` requires the sandbox to be in this process's `running_sandboxes` HashMap, so calling `/snapshot` cold (before any tool call has bound a handle in-process) errors with `sandbox is not running`. The fix is the same `try_resume`-on-miss pattern PR #21 added to `run_in_sandbox`; extending it to `snapshot_sandbox` is a small follow-up orthogonal to this PR.

## What changed

| File | Change |
|---|---|
| `crates/exoharness/src/daytona.rs` (new, ~600 LOC) | REST client + `DaytonaSandboxBackend` + `DaytonaConfig`. Implements `acquire`, `try_resume`, `acquire_from_snapshot`. Synchronous-exec bridge for `start_process`. |
| `crates/exoharness/src/sandbox.rs` | New `SnapshotKind::DaytonaSnapshot` variant + a `(_, DaytonaSnapshot)` arm in the CLI backend's `acquire_from_snapshot` so the wrong-backend case errors clearly. Constants & spec-hash helper exposed `pub(crate)`. |
| `crates/exoharness/src/basic.rs` | `SandboxBackendChoice::Daytona(DaytonaConfig)` variant + `build_sandbox_backend` returns `Result`. |
| `crates/exoharness/src/lib.rs`, `crates/executor/src/lib.rs` | Module registration + re-exports. |
| `crates/cli/src/main.rs` | `--sandbox-backend daytona` + `DaytonaConfig::from_env()` at CLI parse time. |

## Daytona-specific knobs

The REST API has a few non-obvious shapes; documented inline. Highlights surfaced during live debugging:

- `labels` is one JSON-encoded query param, not multiple repeated `?labels=k=v` params.
- List response is `{items, nextCursor}`, not a bare array.
- `env` on create is an object, not an array of `{name, value}`.
- `snapshot` on create refers to a **pre-registered** snapshot name in the user's Daytona dashboard, not an arbitrary docker image ref. We omit it on fresh creates.
- Per-sandbox operations (exec, fs) live on a **different host** — `proxy.app.daytona.io/toolbox/<id>/...` — than the control plane (`app.daytona.io/api`). Both configurable via `DAYTONA_API_URL` / `DAYTONA_TOOLBOX_URL`.
- API keys belong to one organization. Required: `X-Daytona-Organization-ID` header (set via `DAYTONA_ORGANIZATION_ID` env var).

## Out of scope (deliberate)

- **Mounts.** Daytona has no host filesystem to bind-mount. Mounts in `SandboxSpec` are rejected with a clear error. The longer-term answer is a workspace provisioner abstraction at the conversation level — separate effort.
- **Streaming `start_process`.** Daytona's exec endpoint is request/response, not streaming. We buffer stdout/stderr into in-memory `Cursor`s.
- **Cross-backend snapshot interchange.** A `DaytonaSnapshot` payload can't be restored under Docker and vice versa — `kind`-mismatch errors clearly.

## Stack

```
main
  └── #20  sandbox-snapshots-filesystem    Docker snapshot + rewind
        └── #21  cross-process-resume      try_resume contract + 3-tier fallback
              └── #22  this PR             Daytona backend (slots into the contract)
```

## Test plan

- [x] `cargo build --workspace` clean.
- [x] `cargo test --workspace` — 51 unit tests pass; existing integration tests still pass.
- [x] Live end-to-end against real Daytona: `chat send` creates a real sandbox, executes via the toolbox proxy, returns output.
- [x] **Cross-process resume verified live**: two `chat send` invocations in fresh exo processes hit the same Daytona sandbox UUID; marker file written in send 1 is readable in send 2.
- [x] Cleanup via `DELETE /api/sandbox/{id}?force=true` works; no orphaned sandboxes after testing.
- [x] `/snapshot` reaches Daytona and returns a clear error (endpoint not currently deployed upstream — see "Snapshot status" above). Restore path is unreachable until snapshot is available upstream.

### DIFF
--- a/crates/exoharness/src/daytona.rs
+++ b/crates/exoharness/src/daytona.rs
@@ -1,0 +1,610 @@
+//! Daytona remote-container sandbox backend.
+//!
+//! Speaks Daytona's REST API directly via `reqwest`. State persistence is
+//! handled by Daytona itself: `stop` preserves the sandbox filesystem and the
+//! next `acquire` finds the same sandbox by label and `start`s it. Explicit
+//! `/snapshot` and `/rewind` go through [`SnapshotKind::DaytonaSnapshot`]
+//! payloads whose bytes are a small JSON manifest pointing at a named snapshot
+//! in Daytona's registry.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+
+use anyhow::{Context, Result, anyhow, bail};
+use async_trait::async_trait;
+use bytes::Bytes;
+use futures::future::BoxFuture;
+use futures::io::Cursor;
+use reqwest::header::{AUTHORIZATION, CONTENT_TYPE, HeaderMap, HeaderValue};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use crate::sandbox::{
+    DEFAULT_SANDBOX_IMAGE, ManagedSandboxBackend, ManagedSandboxHandle, SandboxCommand,
+    SandboxCommandOutput, SandboxNetworkPolicy, SandboxRequest, SandboxSpec, SnapshotKind,
+    SnapshotPayload, WARM_SANDBOX_KEY_LABEL, WARM_SANDBOX_SPEC_HASH_LABEL, sandbox_spec_hash,
+};
+
+pub const DEFAULT_DAYTONA_API_URL: &str = "https://app.daytona.io/api";
+
+/// Per-sandbox operations (exec, fs) go through Daytona's toolbox proxy rather
+/// than the control-plane API. The two have different base hosts.
+pub const DEFAULT_DAYTONA_TOOLBOX_URL: &str = "https://proxy.app.daytona.io";
+
+#[derive(Debug, Clone)]
+pub struct DaytonaConfig {
+    pub api_key: String,
+    pub api_url: String,
+    pub toolbox_url: String,
+    /// Optional region target (`eu` / `us`). Passed through to Daytona on
+    /// sandbox creation when set.
+    pub target: Option<String>,
+    /// Organization ID to scope requests to. Sent as the
+    /// `X-Daytona-Organization-ID` header. API keys can belong to multiple
+    /// organizations; without this, Daytona may default to a "personal" org
+    /// that isn't the one with credit / the one the user expects.
+    pub organization_id: Option<String>,
+}
+
+impl DaytonaConfig {
+    pub fn from_env() -> Result<Self> {
+        let api_key = std::env::var("DAYTONA_API_KEY")
+            .map_err(|_| anyhow!("DAYTONA_API_KEY is not set; required for the Daytona sandbox backend"))?;
+        let api_url =
+            std::env::var("DAYTONA_API_URL").unwrap_or_else(|_| DEFAULT_DAYTONA_API_URL.to_string());
+        let toolbox_url = std::env::var("DAYTONA_TOOLBOX_URL")
+            .unwrap_or_else(|_| DEFAULT_DAYTONA_TOOLBOX_URL.to_string());
+        let target = std::env::var("DAYTONA_TARGET").ok();
+        let organization_id = std::env::var("DAYTONA_ORGANIZATION_ID").ok();
+        Ok(Self {
+            api_key,
+            api_url,
+            toolbox_url,
+            target,
+            organization_id,
+        })
+    }
+}
+
+/// JSON payload persisted alongside a `SnapshotKind::DaytonaSnapshot`. The
+/// canonical filesystem bytes live in Daytona — we only persist enough
+/// metadata to ask Daytona to create a sandbox from the same snapshot later.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct DaytonaSnapshotManifest {
+    /// Name of the snapshot in Daytona's snapshot registry.
+    snapshot_name: String,
+    /// Image the snapshotted sandbox was originally created from. Used as a
+    /// fallback if the snapshot itself doesn't carry enough info to recreate.
+    base_image: String,
+}
+
+pub struct DaytonaSandboxBackend {
+    client: reqwest::Client,
+    api_url: String,
+    toolbox_url: String,
+    target: Option<String>,
+}
+
+impl DaytonaSandboxBackend {
+    pub fn new(config: DaytonaConfig) -> Result<Self> {
+        let mut headers = HeaderMap::new();
+        let mut auth = HeaderValue::from_str(&format!("Bearer {}", config.api_key))
+            .context("DAYTONA_API_KEY contains characters that aren't valid in an HTTP header")?;
+        auth.set_sensitive(true);
+        headers.insert(AUTHORIZATION, auth);
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
+        if let Some(org_id) = config.organization_id.as_deref() {
+            let value = HeaderValue::from_str(org_id).context(
+                "DAYTONA_ORGANIZATION_ID contains characters that aren't valid in an HTTP header",
+            )?;
+            headers.insert("X-Daytona-Organization-ID", value);
+        }
+        let client = reqwest::Client::builder()
+            .default_headers(headers)
+            .build()
+            .context("building Daytona HTTP client")?;
+        Ok(Self {
+            client,
+            api_url: config.api_url.trim_end_matches('/').to_string(),
+            toolbox_url: config.toolbox_url.trim_end_matches('/').to_string(),
+            target: config.target,
+        })
+    }
+
+    fn api_endpoint(&self, path: &str) -> String {
+        format!("{}{}", self.api_url, path)
+    }
+
+    fn toolbox_endpoint(&self, path: &str) -> String {
+        format!("{}{}", self.toolbox_url, path)
+    }
+
+    async fn find_sandbox_by_labels(
+        &self,
+        key_label: &str,
+        spec_hash: &str,
+    ) -> Result<Option<DaytonaSandbox>> {
+        // Daytona expects `labels` as a single query parameter whose value is a
+        // JSON-encoded object. Both label keys go into the same object so a
+        // spec change yields a fresh sandbox (matching the CLI backend's
+        // "evict on spec change" semantics).
+        let labels_filter = serde_json::json!({
+            WARM_SANDBOX_KEY_LABEL: key_label,
+            WARM_SANDBOX_SPEC_HASH_LABEL: spec_hash,
+        });
+        let response = self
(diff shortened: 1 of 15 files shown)


akrentsel:
Some testing – it's working! 
daytona-resume

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
+//! Daytona remote-container sandbox backend.
```
ankrgyl, line 1: we should probably move this into exoharness/src/sandbox_provider (or something like that). I expect to have like 20 of these and we don't want them to pollute the top-level directory

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+impl DaytonaConfig {
+    pub fn from_env() -> Result<Self> {
+        let api_key = std::env::var("DAYTONA_API_KEY")
+            .map_err(|_| anyhow!("DAYTONA_API_KEY is not set; required for the Daytona sandbox backend"))?;
+        let api_url =
+            std::env::var("DAYTONA_API_URL").unwrap_or_else(|_| DEFAULT_DAYTONA_API_URL.to_string());
+        let toolbox_url = std::env::var("DAYTONA_TOOLBOX_URL")
+            .unwrap_or_else(|_| DEFAULT_DAYTONA_TOOLBOX_URL.to_string());
+        let target = std::env::var("DAYTONA_TARGET").ok();
+        let organization_id = std::env::var("DAYTONA_ORGANIZATION_ID").ok();
```
ankrgyl, line 60: a few principles on arg parsing:

* arguments should always come through clap so they can be provided on the CLI or env
* the exoharness should read 0 environment variables. it eventually becomes a security issue (we want it to be very explicit and trivial to reason about what secrets / config it uses)

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// `X-Daytona-Organization-ID` header. API keys can belong to multiple
+    /// organizations; without this, Daytona may default to a "personal" org
+    /// that isn't the one with credit / the one the user expects.
+    pub organization_id: Option<String>,
+}
+
+impl DaytonaConfig {
+    pub fn from_env() -> Result<Self> {
+        let api_key = std::env::var("DAYTONA_API_KEY")
+            .map_err(|_| anyhow!("DAYTONA_API_KEY is not set; required for the Daytona sandbox backend"))?;
```
ankrgyl: this should be registered as a secret

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+        let mut auth = HeaderValue::from_str(&format!("Bearer {}", config.api_key))
+            .context("DAYTONA_API_KEY contains characters that aren't valid in an HTTP header")?;
+        auth.set_sensitive(true);
+        headers.insert(AUTHORIZATION, auth);
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
+        if let Some(org_id) = config.organization_id.as_deref() {
+            let value = HeaderValue::from_str(org_id).context(
+                "DAYTONA_ORGANIZATION_ID contains characters that aren't valid in an HTTP header",
+            )?;
+            headers.insert("X-Daytona-Organization-ID", value);
```
ankrgyl, line 102: good practice to link to docs from daytona here

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+        }
+        let client = reqwest::Client::builder()
+            .default_headers(headers)
+            .build()
+            .context("building Daytona HTTP client")?;
```
ankrgyl, line 107: not necessary yet but we should eventually factor this in somewhere common because there are a lot of tuning knobs / proxying parameters we'll want to standardize across all http calls (like # of connections in the pool)

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+            client,
+            api_url: config.api_url.trim_end_matches('/').to_string(),
+            toolbox_url: config.toolbox_url.trim_end_matches('/').to_string(),
+            target: config.target,
+        })
+    }
+
+    fn api_endpoint(&self, path: &str) -> String {
+        format!("{}{}", self.api_url, path)
+    }
```
ankrgyl, line 118: i would create a url utils module and add some of this stuff there. also i see you normalize the url above to remove the slash. if i'm reading correctly does that require path starts with a `/`? I think it's a general best practice to strip leading and trailing slashes and then re-insert them when combining path pieces. Every project i work on has a custom urljoin function lol

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+            .query(&[("labels", labels_filter.to_string())])
+            .send()
+            .await
+            .context("listing Daytona sandboxes")?
+            .error_for_status()
+            .context("Daytona list sandboxes returned an error status")?;
+        let list: DaytonaSandboxList = response
+            .json()
+            .await
+            .context("decoding Daytona sandbox list response")?;
```
ankrgyl, line 145: we should either use or create a helper `get_json` type method that we can use across a bunch of APIs

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+fn render_shell_command(argv: &[String]) -> String {
+    // Daytona's execute endpoint takes a single command string interpreted by
+    // a shell. Quote each argv element so spaces / metacharacters survive.
+    argv.iter()
+        .map(|arg| shell_quote(arg))
+        .collect::<Vec<_>>()
+        .join(" ")
+}
+
+fn shell_quote(arg: &str) -> String {
```
ankrgyl, line 520: this should prob also be in a generic utils module

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+#[derive(Debug, Deserialize)]
+struct DaytonaSandboxList {
+    #[serde(default)]
+    items: Vec<DaytonaSandbox>,
+}
+
+#[derive(Debug, Deserialize)]
+struct DaytonaSandbox {
+    id: String,
+    state: String,
```
ankrgyl, line 588: would be better to have an enum of known states here if it's documented by daytona.

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,610 @@
...
+        } else {
+            quoted.push(c);
+        }
+    }
+    quoted.push('\'');
+    quoted
+}
+
+#[derive(Debug, Serialize)]
+struct DaytonaCreateRequest {
```
ankrgyl, line 542: i want to be careful about supply chain stuff and going crazy with dependencies, but people have created daytona SDKs (https://github.com/krzysztofwos/daytona-client) and they also publish an openapi spec (https://www.daytona.io/docs/en/tools/api/#daytona/tag/config) which we could auto generate types from

crates/exoharness/src/daytona.rs
```diff
@@ -0,0 +1,602 @@
...
+/// canonical filesystem bytes live in Daytona — we only persist enough
+/// metadata to ask Daytona to create a sandbox from the same snapshot later.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct DaytonaSnapshotManifest {
+    /// Name of the snapshot in Daytona's snapshot registry.
+    snapshot_name: String,
+    /// Image the snapshotted sandbox was originally created from. Used as a
+    /// fallback if the snapshot itself doesn't carry enough info to recreate.
+    base_image: String,
+}
```
Bautista1999, line 81: Random (and dumb) question: If Daytona already persists the filesystem in the snapshot, why do we need the `base_image`? Wouldn't the snapshot_name or id be enough to get the snapshot back on a new sandbox?

akrentsel:
Thanks for the comments @ankrgyl. I just went ahead and re-wrote this afresh (taking your comments into account, and your newly-merged #25). It was different enough that I figured I'd have a new PR for it, and I'll close this one.

#45 is the new one. I'm going to follow thi sup with some changes to adapt Juan's code to fit this pattern too.

PR #23, opened 2026-05-25
Resolve model wire format per model instead of hardcoding Responses

Alexsun1one:
## Summary

The executor hardcoded `ProviderFormat::Responses` for every model, so anything that doesn't speak the OpenAI Responses API failed — DeepSeek and other OpenAI-compatible providers 404 on `/responses`. (It also meant streaming from chat-completions models was parsed with the wrong format.)

Resolve the wire format per model instead:

- When `EXO_MODEL_LIST_PATH` points at a Braintrust-style `model_list.json`, the model's format/flavor come from there.
- Otherwise the model is resolved by name via lingua's built-in detection, defaulting unknown / self-hosted models to Chat Completions (what OpenAI, DeepSeek, Together, Groq, vLLM, Ollama, ... all speak).
- OpenAI's Responses-only families (o1-pro, o3-pro, gpt-5-pro, gpt-5-codex, gpt-5.3+) still route to the Responses API. The registry encodes them as `openai`/`chat` and distinguishes them by name, so Chat Completions is upgraded to Responses through the public `ModelSpec::requires_responses_api()`.

No substrate changes; this is entirely in the executor.

## On using a public model registry

The catalog is loaded explicitly (consistent with how `braintrust-llm-router` is designed — it ships no bundled default), via `EXO_MODEL_LIST_PATH`. Name-based detection is the zero-config default, so the common OpenAI-compatible cases work with no setup.

## Scope / follow-ups

- This fixes the wire *format*. Native Anthropic/Google/Bedrock transport (different auth + endpoints) is a natural follow-up — the format already resolves correctly for those, but requests still go through the OpenAI-compatible provider.
- Only the Basic and Rlm harnesses use this path; the TypeScript harness dispatches models in JS.

## Testing

- Unit tests: Responses-only upgrade by name, Chat Completions default for unknown models, registry-format passthrough, max-output-tokens fallback.
- `cargo test --workspace --all-targets` is green.

### DIFF
--- a/crates/executor/src/harness_runtime.rs
+++ b/crates/executor/src/harness_runtime.rs
@@ -32,12 +32,11 @@ pub struct RouterModelClient {
 }
 
 impl RouterModelClient {
-    pub fn new(env: HashMap<String, String>) -> Result<Self> {
-        let catalog = load_model_catalog(&env)?;
-        Ok(Self {
+    pub async fn new(env: HashMap<String, String>) -> Self {
+        Self {
             env: Arc::new(env),
-            catalog: Arc::new(catalog),
-        })
+            catalog: shared_model_catalog().await,
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/executor/src/harness_runtime.rs
```diff
@@ -144,8 +149,86 @@ fn optional_env(env: &HashMap<String, String>, key: &str) -> Option<String> {
...
+/// capabilities (wire format, flavor, context window, ...). When unset, models
+/// are resolved by name via lingua's built-in model-family detection.
+const MODEL_LIST_PATH_ENV: &str = "EXO_MODEL_LIST_PATH";
+
+fn load_model_catalog(env: &HashMap<String, String>) -> Result<ModelCatalog> {
+    match optional_env(env, MODEL_LIST_PATH_ENV) {
+        Some(path) => Ok(ModelCatalog::from_file(&path)?),
+        None => Ok(ModelCatalog::empty()),
+    }
+}
```
ankrgyl: i think probably better to load this once at startup time using a github link

Alexsun1one: ok. switched it to pull the list from the github raw link once at startup. falls back to name-based detection if the fetch fails so it still works offline. is this ok？

akrentsel:
Stale – closing (but feel free to re-open if this is still needed)

PR #25, opened 2026-05-25
HTTP protocol support

ankrgyl:
You can run

```
exo serve -v
```

from one place and

```
exo --exoharness-url http://127.0.0.1:4766 repl
```

elsewhere, which will use the server.

## Follow-up TODOs

- Basic ExoHarness: avoid reconstructing stateless turn handles by scanning the full conversation history. A reasonable implementation is to maintain a per-turn index, e.g. `conversations/<conversation_id>/turns/<turn_id>/record.json` plus optional per-turn event copies, so `turn_handle({ session_id, turn_id })` is O(1) or O(turn history) instead of O(conversation history).
- Hosted/remote ExoHarness: make sandbox process output/status durable instead of keeping it only in live process memory; the durable-looking sandbox process event variants should either be written to the event log or removed until the durable implementation lands.


crates/cli/src/main.rs
```diff
@@ -174,12 +194,21 @@ enum SecretBackendArg {
...
 #[derive(Debug, Clone, Copy, ValueEnum)]
-enum SandboxBackendArg {
+enum SandboxProviderArg {
+    Daytona,
     #[value(name = "apple-container")]
     AppleContainer,
     Docker,
-    #[value(name = "local-process")]
-    LocalProcess,
+}
```
ankrgyl, line 206: re-add local process?

crates/cli/src/main.rs
```diff
@@ -1916,18 +2346,26 @@ fn slugify(input: &str) -> String {
...
         );
     }
 
     loaded_env
         .get(env)
         .cloned()
         .or_else(|| std::env::var(env).ok())
-        .ok_or_else(|| anyhow!("environment variable passed to --env is not set"))
+        .ok_or_else(|| anyhow!("environment variable passed to {flag} is not set"))
 }
```
ankrgyl, line 2399: not sure this function should exist. why can't it just use clap?

crates/cli/src/main.rs
```diff
@@ -484,10 +623,35 @@ async fn main() -> Result<()> {
...
+    }
+
+    if cli.bearer_env.is_some() && cli.exoharness_url.is_none() {
+        bail!("--bearer-env requires --url");
+    }
+    let bearer_token = cli
+        .bearer_env
+        .as_deref()
+        .map(|env| env_value_from_arg("--bearer-env", env, &env_vars))
+        .transpose()?;
```
ankrgyl, line 640: why not use clap?

crates/cli/src/main.rs
```diff
@@ -1405,17 +1644,82 @@ fn command_agent_ref(command: &Commands) -> Option<&str> {
...
+    if !config.bind.ip().is_loopback() {
+        anyhow::bail!(
+            "exo serve only binds loopback addresses; got {}",
+            config.bind
+        );
+    }
+    let exoharness = Arc::new(BasicExoHarness::new(exo_config.clone()).await?);
+    let listener = TcpListener::bind(config.bind)?;
+    let addr = listener.local_addr()?;
+    println!("serving exoharness HTTP on http://{addr}");
```
ankrgyl: should be an info level tracing statement

PR #28, opened 2026-05-27
Exoclaw + Task Scheduling + Adapters

61cygni:
Adds Exoclaw, a long-running local agent example built on the TypeScript harness, with support for scheduled sandbox tasks and host-managed external adapters. I tried as much as possible to keep changes to examples/exoclaw/. PR is relatively large because I had to add task support and adapter support, modify sandbox support for different lifespan and scoping. 

- Adds Exoclaw setup, prompts, docs, and launcher script for running the REPL, scheduler runner, and adapter runner together.
- Adds scheduled task storage/runtime support so Exoclaw can create recurring sandbox tasks and wake the conversation with run results.
- Adds the adapter runtime and tools for long-running IRC, WhatsApp, and Signal workers, including inbound wakeups and explicit outbound replies.

Manually tested fresh Exoclaw setup with Signal, WhatsApp, IRC, scheduled tasks, and adapter replies


examples/typescript/turn-loop.ts
```diff
@@ -0,0 +1,178 @@
...
+    context: TurnContext,
+  ) => Promise<void> | void;
+}
+
+export async function runResponsesHarnessTurn(
+  context: TurnContext,
+  options: ResponsesTurnLoopOptions = {},
+): Promise<void> {
+  const modelBinding = await resolveLlmBinding(context);
+  const runtime = ResponsesRuntime.fromModelBinding(
```
Alexsun1one: `ResponsesRuntime.fromModelBinding(...)` called directly, no dispatcher. `typescript/model-runtime/` only has `responses.ts`. any non-Responses provider 404s on `/responses`.

PR #23 fixed this on the rust side; the TS path is the same hardcoding.

fix: add a chat-completions runtime sibling and dispatch per model via `requiresResponsesApi(model)`. providers like deepseek also reject the `developer` role so the chat-completions path needs developer→system mapping.

examples/typescript/turn-loop.ts
```diff
@@ -0,0 +1,178 @@
...
+          parent: turnParent,
+          roundIndex: round,
+        });
+
+    const events = responseToLinguaEvents(response);
+    if (events.length > 0) {
+      latestEventId = await appendTurnEvents(context, events);
+    }
+
+    const toolCalls = responseToolCalls(response);
```
Alexsun1one, line 150: `responseToolCalls(response)` called without try/catch. `parseJsonObject` inside `responses.ts` wraps `JSON.parse` with no error handling. when the model emits malformed JSON in long tool arguments (unterminated string in multi-thousand-char `arguments`), the throw escapes and exits the harness non-zero, taking the REPL with it.

```
SyntaxError: Unterminated string in JSON at position 11232
  at parseJsonObject (model-runtime/responses.ts:597)
  at responseToolCalls (responses.ts:476)
  at runResponsesTurnLoop (turn-loop.ts)
```

fix: `parseJsonObject` returns a discriminated result; `responseToolCalls` emits a synthetic `tool_result` error per malformed call so the next round lets the model self-correct. standard agent-loop behavior.

examples/exoclaw/adapters/signal/worker.ts
```diff
@@ -0,0 +1,567 @@
...
+  throw new Error("Signal trigger must be all_messages or contacts_only");
+}
+
+await fs.mkdir(configDir, { recursive: true });
+const account = configuredAccount ?? (await discoverOrLinkAccount());
+const signal = spawnSignalCli([
+  "-a",
+  account,
+  "jsonRpc",
+  "--receive-mode=on-connection",
```
Alexsun1one: `--receive-mode=on-connection` only triggers receive when a JSON-RPC client connects over a socket. this worker uses stdio pipes, so signal-cli never starts receiving. outbound `send` works, inbound `dataMessage` envelopes never reach the worker.

fix: `--receive-mode=on-start`.

examples/exoclaw/adapters/whatsapp/worker.ts
```diff
@@ -0,0 +1,221 @@
...
+  logger,
+  printQRInTerminal: false,
+  syncFullHistory: false,
+  version,
+});
+
+socket.ev.on("creds.update", saveCreds);
+
+socket.ev.on("connection.update", (update) => {
+  if (update.qr) {
```
Alexsun1one: QR-only path. on a china-routed proxy, Baileys 7.0.0-rc13 WebSocket dies 30-80s into the pairing handshake; the post-scan key exchange cant complete and the phone reports `unable to link device`.

fix: add Baileys' `requestPairingCode(<phone>)` flow as an alternative to QR. it runs over HTTPS rather than the persistent WS so it isnt killed mid-handshake. config flag like `linkMethod: "qr" | "pairing-code"`, emit a `pairing_code` lifecycle event analogous to `link_qr`.

Alexsun1one (commented):
findings from testing origin/adapters yesterday. items below still reproduce on the PR head — they're worker.ts / harness-layer behavior not the rust restructure. 4 inline notes anchored to specific lines, 3 here without a clean single-line anchor.

## `--env-file-if-exists` vars dont reach worker subprocesses

`crates/cli/src/env.rs::CliEnvironment::load` parses .env into a `HashMap<String, String>` and returns via `into_vars()` without calling `std::env::set_var`. when the adapter runtime spawns workers via `Command::new(...).env(...)`, the child only sees `EXO_*` + secretEnv + parent shell env. .env-only vars are invisible to the worker.

repro: a feishu adapter reading `process.env.FEISHU_APP_ID` died with `Feishu adapter needs an app id` even though the var was in .env and exo was launched with `--env-file-if-exists .env`.

fix: call `std::env::set_var` on each loaded entry, or document .env as exo-internal and require workers to use secretEnv. current behavior is a footgun — .env looks like standard dotenv but isnt.

## `apple-container` default sandbox backend fails with generic ENOENT on macs without it

`conversation send` returns `Error: No such file or directory (os error 2)` within 2s on fresh macs because the default backend is apple-container and the `container` CLI isnt installed.

fix: detect `container` on PATH and fall back to local-process with a warning, or replace the ENOENT with a backend-specific error naming the missing binary.

workaround: `--sandbox-backend local-process`.

## lingua TS materialization breaks the OpenAI `tool_calls` adjacency contract

when an assistant turn contains N `tool_calls`, lingua's TS-side materialization fails to keep the N corresponding `tool` messages adjacent. the follow-up chat-completions request 400s on providers that enforce this (deepseek and others).

repro: 3 `disable_adapter` calls in one assistant turn, the follow-up 400s. lingua-ts bug surfacing in exo's turn loop.

Alexsun1one:
two retractions on my earlier review (#4374207281), on re-read:

**re: .env not reaching workers (review body)** - on second thought .env not reaching workers is probably by design, workers should use secretEnv explicitly. the gap is the cli flag name doesnt hint at this. so its a doc issue not a bug. ignore the original framing.

**re: lingua tool_calls adjacency (review body)** - actually this one is out of scope. lingua materialization is a lingua bug, just surfaces in exo's turn loop. should be in the lingua repo not here. sorry for the noise.

examples/typescript/turn-loop.ts
Alexsun1one: fwiw examples/typescript/turn-loop.ts reads like a demo but examples/exoclaw/harness.ts imports runResponsesTurnLoop from it directly, so its on the live path for exoclaw users not just example code. the Responses-only hardcoding bites whenever someone points exoclaw at deepseek or similar non-Responses provider.

examples/exoclaw/adapters/signal/worker.ts
Alexsun1one: evidence i forgot to attach. on receive-mode=on-connection: signal-cli child idle on receive side, outbound send works fine, zero inbound dataMessage envelopes for ~4h session on origin/adapters. switched to on-start locally, inbound started flowing within seconds. can paste scrubbed log if useful.

crates/executor/src/conversation_wakeup.rs
```diff
@@ -0,0 +1,21 @@
...
+use lingua::universal::UserContent;
+
+use crate::{HarnessConversation, SendRequest, SendResult};
+
+pub async fn send_conversation_wakeup(
+    conversation: &dyn HarnessConversation,
+    prompt: String,
+) -> Result<SendResult> {
+    let result = conversation
+        .send(SendRequest {
```
Alexsun1one, line 20: this is the core of the contradiction in the review body. every wakeup opens a fresh session (session_id: None) and the conversation.send call advances head, so any in-flight turn on the same conversation goes stale.

the `runtime.rs` inline comments on L230/L268 acknowledge the danger but the fix walks into it via this call. serialize (queue behind active turn) or partition (system-scoped conversation) are the two shapes that preserve `ensure_conversation_head`'s single-writer invariant.

crates/executor/src/adapter/store.rs
```diff
@@ -0,0 +1,352 @@
...
+        match fs::read(&path).await {
+            Ok(bytes) => Ok(Some(serde_json::from_slice::<AdapterRecord>(&bytes)?)),
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => Ok(None),
+            Err(error) => {
+                Err(error).with_context(|| format!("failed to read adapter {}", path.display()))
+            }
+        }
+    }
+
+    pub async fn put_adapter(&self, adapter: &AdapterRecord) -> Result<()> {
```
Alexsun1one, line 98: `put_adapter` reads the json file, mutates the in-memory record, writes back without flock or atomic rename. same pattern at `mark_connected` (L130), `mark_error` (L144), `disable_adapter` (L103).

concrete race: scheduler-runner fires a task that triggers `mark_connected('adapter-x')` while adapter-runtime is mid-`mark_error('adapter-x')` for the same adapter. last writer wins, the lost write's state never surfaces. on a json file this is also a torn-write risk if the process is killed mid `fs::write`.

fix: write-to-temp + `fs::rename` for atomic replace, or tokio Mutex per adapter id at the store layer.

crates/executor/src/adapter/store.rs
```diff
@@ -0,0 +1,352 @@
...
+        while let Some(entry) = entries.next_entry().await? {
+            let path = entry.path();
+            if path.extension().and_then(|ext| ext.to_str()) != Some("json") {
+                continue;
+            }
+            let bytes = fs::read(&path).await.with_context(|| {
+                format!("failed to read adapter outbound message {}", path.display())
+            })?;
+            let message = serde_json::from_slice::<AdapterOutboundMessageRecord>(&bytes)?;
+            remove_file_if_exists(path).await?;
```
Alexsun1one: read-then-delete is not atomic. two failure modes:

1. process crash between L212 `fs::read` and L216 `remove_file_if_exists`: same message gets re-read and double-delivered next time `take_outbound_messages` is called. for whatsapp/signal adapters this means the user sees the same inbound message twice in their conversation.
2. two readers race (e.g. adapter-runtime restart while scheduler-runner also drains): both pass the read step, both pass the delete (idempotent), both deliver the same message downstream.

fix: rename-then-read (move to a `delivered/` subdir before processing) so the read step is naturally idempotent.

crates/executor/src/conversation_sandbox.rs
```diff
@@ -0,0 +1,142 @@
...
+) -> Result<String> {
+    let spec = conversation_sandbox_spec(agent_config, config);
+
+    for sandbox in conversation_sandboxes(conversation).await? {
+        if sandbox.matches_spec(&spec) {
+            return Ok(sandbox.id);
+        }
+    }
+
+    create_conversation_sandbox(conversation, agent_config, config).await
```
Alexsun1one, line 58: check-then-create race. two callers concurrent on a cold conversation:

1. caller A enters `ensure_conversation_sandbox`, `conversation_sandboxes()` returns empty.
2. caller B enters the same function, `conversation_sandboxes()` also returns empty.
3. both fall through to `create_conversation_sandbox`, both succeed.
4. conversation now has two `sandbox_created` events with the same spec, the second container is wasted.

on `apple-container` backend this means 2x container processes for one conversation, harder to attribute resource use and slower cold start because of duplicate image pulls.

caller A could be the REPL's first turn, caller B could be a scheduled task firing in the cold-start window or an adapter wakeup arriving from inbound traffic. odds go up with `@every 10s` tasks or chatty adapters like signal `all_messages`.

fix: per-conversation lock around the (check, create) pair, or rely on `SandboxCreated` event idempotency by passing a deterministic key derived from the spec.

typescript/harness/adapter-tools.ts
```diff
@@ -0,0 +1,486 @@
...
+  }
+}
+
+function transformAdapterConfig(config: JsonObject): JsonObject {
+  const type = stringField(config, "type");
+  if (type === "irc") {
+    const passwordSecretId = nullableStringField(config, "passwordSecretId");
+    return {
+      adapterType: "irc",
+      workerCommand: ["pnpm", "tsx", "examples/exoclaw/adapters/irc/worker.ts"],
```
Alexsun1one: `workerCommand: ["pnpm", "tsx", "examples/exoclaw/adapters/irc/worker.ts"]` is a path relative to process CWD. works when exo is launched from the repo root, breaks when launched from anywhere else.

same pattern at L358 (whatsapp) and L375 (signal).

`.exo/exoclaw-profile.md` had the same issue from a different surface (already a setup gotcha for users who symlink the binary or run `exo` from a project subdir). worth fixing both at the same source-of-truth: resolve worker path relative to a known anchor (the harness module dir / `import.meta.url`) rather than CWD.

Alexsun1one (commented):
follow-up to my earlier review. one substrate-level item that didnt fit inline, plus a store-layer concurrency pattern i noticed reading the rust side. 5 inline notes anchored to specific lines.

## turn-stale spec contradicts the scheduler/adapter wakeup paths

`crates/exoharness/src/basic.rs::ensure_conversation_head` enforces strict turn-owned head: while a turn is in flight, any external write to the same conversation makes the turn unresumable. the test `stale_turn_artifact_write_reports_unresumable_turn` at `crates/exoharness/src/basic_tests.rs:178` confirms this is intended. it deliberately writes an artifact from outside the turn and asserts the in-flight turn fails with `turn is stale and cannot be resumed`.

but this PR has multiple writers to the same conversation by design:

- `crates/executor/src/conversation_wakeup.rs::send_conversation_wakeup` calls `conversation.send(SendRequest { input: vec![user_message], session_id: None })` on each scheduled task fire. that opens a new session and advances head.
- `examples/exoclaw/adapter-architecture.md` second paragraph: "A scheduled task ... writes output, wakes the conversation, and exits."
- the mermaid diagram in the same file shows both `scheduler -> convo` and `runtime -> wakeup -> convo`.

so scheduler-runner + adapter-runtime are designed as legitimate writers, but any concurrent user turn on that conversation fails. concrete from yesterday on origin/adapters (same commit as this PR head `599aed25`):

- user turn started `04:57:18.926Z`
- every-1m scheduler task fired in the gap, advanced head to `04:57:20.788Z`, 1.8s
- user turn next `add_events` failed with `turn is stale and cannot be resumed: conversation head advanced outside this turn`

any user turn that takes more than the cron tick interval is unsafe on a conversation that also has scheduled tasks or active adapters. with a single `@every 1m` task, any user turn longer than ~60s is at risk. with `@every 10s` (or an active adapter relaying inbound messages), the risk window shrinks to seconds.

the docs say a scheduled task "wakes the conversation" but the conversation has a single-writer invariant. these cant both hold under load.

i noticed you flagged the same risk in 3 inline comments on `crates/executor/src/adapter/runtime.rs` (L88, L230, L268). but the runtime.rs fix path defers to `send_conversation_wakeup`, which does exactly the head-advance that triggers the invariant. so those comments are right about the danger but the implementation still walks into it on the wakeup path.

two substrate-level fix shapes:

1. **serialize**: scheduler/adapter wakeups queue and wait for the in-flight turn to finish before opening a new session on the same conversation. preserves the strict turn-owned head invariant, eliminates user-visible stale errors.
2. **partition**: scheduler/adapter writes land on a separate (system-scoped) conversation, and the user conversation only consumes wakeup signals via a queue rather than as direct conversation.send writes. strict head stays correct on user conversations.

## store-layer concurrency, smaller theme

the file-based stores added in this PR (`AdapterStore`, `SchedulerStore`) do read-modify-write without flock or atomic rename. with scheduler-runner + adapter-runtime + the main REPL all reaching into `$root/.exo/adapters/*.json` or `$root/.exo/scheduled-tasks/tasks/*.json`, two near-simultaneous writes can lose updates. inline notes on the specific call sites plus a related check-then-create race on `ensure_conversation_sandbox` and a config-path footgun on the adapter worker commands.

if either turn-stale direction works, i can prototype.

crates/cli/src/adapters.rs
```diff
@@ -0,0 +1,117 @@
...
+                .await?
+                .into_iter()
+                .filter(|adapter| include_disabled || adapter.enabled)
+            {
+                println!(
+                    "{}\t{}\t{:?}\t{}",
+                    adapter.id, adapter.enabled, adapter.source, adapter.name
+                );
+            }
+        }
```
ankrgyl, line 53: use the new table formatting library here

crates/cli/src/adapters.rs
```diff
@@ -0,0 +1,117 @@
...
+                bail!("adapter not found: {adapter_id}");
+            }
+        }
+    }
+    Ok(())
+}
+
+struct AdapterRunnerLock {
+    path: std::path::PathBuf,
+}
```
ankrgyl, line 79: does this only need to be locked on a single machine (eg in a sandbox) or require global coordination?

crates/cli/src/adapters.rs
```diff
@@ -0,0 +1,117 @@
...
+                let existing_pid = fs::read_to_string(&path).unwrap_or_default();
+                if process_is_running(existing_pid.trim()) {
+                    bail!(
+                        "adapter runner already appears to be running with pid {}",
+                        existing_pid.trim()
+                    );
+                }
+                fs::remove_file(&path)?;
+                Self::acquire(root)
+            }
```
ankrgyl, line 105: would have a coding agent write a good stress test for this

crates/executor/src/adapter/runtime.rs
```diff
@@ -0,0 +1,356 @@
...
+                        Ok(_) => break,
+                        Err(_) => break,
+                    }
+                    if let Err(error) =
+                        run_adapter_loop(Arc::clone(&harness), &store, adapter.clone()).await
+                    {
+                        eprintln!(
+                            "adapter {} runtime error: {error}; restarting in 5s",
+                            adapter.id
+                        );
```
ankrgyl: use tracing

ankrgyl: i can do a big pass over the code now and replace it all though separately

crates/executor/src/adapter/runtime.rs
```diff
@@ -0,0 +1,356 @@
...
+                        let _ = store.mark_error(&adapter.id, error.to_string()).await;
+                        let _ = store
```
ankrgyl: `let _` is a last report and usually means you are swallowing an error. if tehre's an error and you want to ginore it, i'd suggest `trace::warn`ing it at the very least

crates/executor/src/adapter/store.rs
```diff
@@ -0,0 +1,370 @@
...
+use tokio::fs;
+
+use super::types::{
+    AdapterAttachment, AdapterEventRecord, AdapterEventType, AdapterOutboundMessageRecord,
+    AdapterRecord, NewAdapter, now_ms,
+};
+
+#[derive(Debug, Clone)]
+pub struct AdapterStore {
+    root: PathBuf,
```
ankrgyl, line 16: better to use object store unless this is guaranteed to only run locally

ankrgyl: (object store writes locally too. but can also write in s3, etc)

ankrgyl (commented):
generally lgtm

PR #29, opened 2026-05-27
cli: /usage (alias /cost) REPL command

akrentsel:
## Summary

Adds a `/usage` command (alias `/cost`) to the chat REPL so you can see, at any point in a conversation, what it has cost so far. This is the demo surface for the per-message cost tracking in #16.

Example output:

```
usage (this conversation):
  model calls : 3
  input       : 14,500 tokens (1,200 cached, 800 cache-write)
  output      : 2,100 tokens (450 reasoning)
  cost        : $0.0182
  model time  : 8.3s
```

If some calls used a model not in the LiteLLM price table, their tokens still count but they're excluded from the dollar total, with an honest note:

```
  note        : 1 of 3 call(s) had no price in the table; cost is a partial total
```

## How it works

- New `/usage` / `/cost` branch in the REPL input loop (`tui.rs`), alongside the existing `/history`, `/quit`, `/exit`.
- `print_usage` reads the conversation's `messages` events via `get_events` and folds the `UsageRecord` on each — no new persistence, it's pure read-side.
- Aggregation/formatting split into pure functions (`summarize_usage`, `render_usage`, `with_commas`) so they're unit-testable without a live model.
- `priced_calls` is tracked separately from `calls` so the partial-total case is reported rather than silently undercounted.

## Stacking

**Stacked on #16** (`feature/message-cost-tracking`) — this reads the `UsageRecord` that #16 introduces, so its base is that branch, not `main`. Will retarget to `main` once #16 merges. Also adds a one-line re-export of `UsageRecord` from `executor` for the CLI.

## Test plan

- [x] `cargo test --workspace` — +4 unit tests in `tui::tests`:
  - `summarize_usage_aggregates_priced_calls_and_skips_usageless_events` (user-input events without usage are ignored; totals sum correctly)
  - `summarize_usage_counts_unpriced_calls_separately` (None cost → counted but not priced; partial-total note rendered)
  - `render_usage_reports_empty_conversation`
  - `with_commas_groups_thousands`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`

## Not in scope

- Per-turn breakdown / live inline display after each turn (this is a cumulative on-demand summary).
- A REPL command help banner (none exists today; `/usage` follows the same undocumented-but-discoverable convention as `/history`).

### DIFF
--- a/crates/cli/src/tui.rs
+++ b/crates/cli/src/tui.rs
@@ -364,6 +364,24 @@ impl ChatRepl {
         Ok(())
     }
 
+    async fn print_usage(&self) -> Result<()> {
+        let result = self
+            .conversation
+            .exoharness_handle()
+            .get_events(Some(EventQuery {
+                cursor: None,
+                direction: Some(EventQueryDirection::Asc),
+                limit: None,
+                session_id: None,
+                turn_id: None,
+                types: Some(vec!["messages".to_string()]),
+            }))
+            .await?;
+        let summary = summarize_usage(result.events.iter().map(|event| &event.data));
+        print!("{}", render_usage(&summary));
+        Ok(())
+    }
+
     async fn run(&mut self) -> Result<()> {
         loop {
             let prompt = format!("{}> ", self.conversation.record().slug);
@@ -378,6 +396,10 @@ impl ChatRepl {
                     }
                     if trimmed == "/history" {
                         self.print_transcript().await?;
+                        continue;
+                    }
+                    if trimmed == "/usage" || trimmed == "/cost" {
+                        self.print_usage().await?;
                         continue;
                     }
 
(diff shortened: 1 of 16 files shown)


crates/cli/src/tui.rs
```diff
@@ -493,6 +515,115 @@ fn print_ttft(ttft: Duration) {
     }
 }
 
+/// Running totals over the `UsageRecord`s attached to a conversation's
+/// model responses. `priced_calls` tracks how many of `calls` carried a
+/// `cost_usd` — so a partial total (some models missing from the price
+/// table) can be reported honestly rather than silently undercounting.
+#[derive(Debug, Default, PartialEq)]
+struct UsageSummary {
```
ankrgyl, line 525: why not just reuse the const struct from Lingua? The only difference appears to be the priced_calls counter which seems like it could be derived, or you could make a composite that embeds the lingua struct?

Pushing on this a bit because usage definitions are changing rapidly and getting more complicated (Eg anthropic recently introduced tiers of cached tokens)

crates/cli/src/tui.rs
```diff
@@ -493,6 +515,115 @@ fn print_ttft(ttft: Duration) {
...
+        ));
+    }
+
+    let mut lines = vec![
+        "usage (this conversation):".to_string(),
+        format!("  model calls : {}", summary.calls),
+        format!("  input       : {input}"),
+        format!("  output      : {output}"),
+        format!("  cost        : ${:.4}", summary.cost_usd),
+        format!(
```
Alexsun1one, line 607: `${:.4}` looks fine for the example values but rounds small costs to misleading shapes:

- a cached call ending in $0.00004 prints `cost : $0.0000`, which reads as free
- a $0.00005001 call prints `$0.0001`, off by ~2x in display terms

few options if it's worth caring about:
- dynamic precision: 4 digits when >= $0.001, 6 digits below
- swap the unit when total < $0.01 (e.g. `42.5 ¢` or `5,432 micro-USD`)
- print exact `{:e}` for very small totals

not blocking, just flagging because the partial-total note already shows that getting cost honest matters here.

Alexsun1one (commented):
clean read-side feature, the priced_calls / partial-total note is a nice touch for honesty. one small UX thing inline on the cost formatting.

crates/cli/src/tui.rs
akrentsel: thanks, addresse

crates/cli/src/tui.rs
akrentsel: yup, that makes sense, can just embed it here + add costs. done.

akrentsel:
updated to address both comments

ankrgyl (approved):

akrentsel:
this got included in a subsequent PR, closing as not relevant anymore.

Issue #30, opened 2026-05-27
`exo repl --model X` is silently ignored when the agent already exists


akrentsel:
## Summary

`exo repl --model <model>` only takes effect when it *creates* a new agent. If an agent with the target slug already exists, the REPL reuses it as-is and the `--model` flag is silently ignored — no warning, no error. The session keeps using whatever model the agent was first created with.

## Repro

```sh
exo repl --model gpt-4o-mini   # creates the default agent pinned to gpt-4o-mini
# ...quit...
exo repl --model gpt-5.5       # silently still uses gpt-4o-mini
```

`/usage` (or any model call) confirms the older model is still in use.

## Expected

One of:
- Honor `--model` for the session (per-session override), or
- Update the agent's model, or
- At minimum, print a warning that `--model` is being ignored because the agent already exists.

## Actual

`--model` is silently dropped; the pre-existing agent's model is used with no feedback.

## Where

`crates/cli/src/main.rs`, the `Commands::Repl` arm. When `harness.get_agent(&agent_slug)` returns `Some`, the agent is returned without consulting the `model` argument (only a harness-selection compatibility check runs). The `model` value is used only in the `None` (create) branch.

## Notes

Low severity (workaround: use `--agent <fresh-slug>` or a clean `--root`), but it's a real least-surprise footgun — easy to think you've switched models when you haven't.

Issue #24, opened 2026-05-25
OpenAI Responses API: cache misses on every request (no prompt_cache_key sent)


akrentsel:
## Summary

Requests issued to OpenAI's `/v1/responses` endpoint through exo never hit
OpenAI's prompt cache, even when the same long prefix is sent repeatedly
within the cache TTL. Users on gpt-4o-family (and other Responses-API-capable
models) pay full input rate instead of the 50% cached-input discount that
should be available automatically.

Noticed while developing #16 (per-message cost / token / latency tracking) —
the new `UsageRecord` infrastructure correctly persisted what OpenAI sent
back, but `prompt_cached_tokens` was always `0`, which turned out to surface
this upstream issue rather than a PR16 bug.

## Reproduction

```bash
# Set up a fresh exo root with an OpenAI binding.
export EXO_SECRET_BACKEND=file
export EXO_SANDBOX_BACKEND=local-process
export EXO_MASTER_KEY_PATH=$PWD/master.key
exo --root .exo secret set openai-key --value "$OPENAI_API_KEY"
exo --root .exo model register gpt-4o-mini --model gpt-4o-mini --secret openai-key
exo --root .exo agent create demo --model gpt-4o-mini

# Build a stable prompt over the 1024-token caching minimum.
python3 -c "print(('The quick brown fox jumps over the lazy dog. ' * 250).strip())" > padding.txt
PROMPT="$(cat padding.txt) Reply with one short sentence."

# Send the same prompt to three fresh conversations.
for c in cache-a cache-b cache-c; do
  exo --root .exo conversation create demo "$c"
  exo --root .exo chat send demo "$c" "$PROMPT" >/dev/null
done

# Inspect the usage records.
for c in cache-a cache-b cache-c; do
  echo "=== $c ==="
  exo --root .exo conversation events demo "$c" --type messages \
    | jq '.events[] | select(.data.usage != null) | .data.usage'
done
```

Observed: all three calls report `prompt_cached_tokens: 0` despite a
~2551-token shared prefix sent within a few seconds of each other.

## What I confirmed it isn't

**Not a lingua parsing bug.** `crates/lingua/src/universal/response.rs:321-324`
reads `input_tokens_details.cached_tokens` for `ProviderFormat::Responses`,
which is the correct OpenAI field. When OpenAI returns a non-zero value,
lingua passes it through to `UniversalUsage.prompt_cached_tokens` (verified by
direct curl tests, see below).

[...]

PR #31, opened 2026-05-27
fix: honor repl --model (#30) + send prompt_cache_key on Responses API (#24)

akrentsel:
Fixes two bugs found while building the cost-tracking stack. **Stacked on #29** (`usage-command`), since the before/after is demonstrated via that PR's `/usage` command. Two independent commits.

## #30 — `exo repl --model X` silently ignored on an existing agent

`exo repl` reused an existing agent and dropped `--model` with no feedback, so you'd think you'd switched models but hadn't. Now an explicit `--model` that differs from the agent's stored model **updates it and prints a notice**:

```
updated agent 'demo' model: gpt-4o-mini -> gpt-5.5
```

No-op when it already matches. (`crates/cli/src/main.rs`, the `Repl` arm.)

## #24 — no `prompt_cache_key` on the OpenAI Responses API → ~0% cache hits

The Responses endpoint needs a stable `prompt_cache_key` to reliably route prompt-cache hits at low request volume; exo never sent one, so cached input was billed at full rate and `UsageRecord.prompt_cached_tokens` was always 0.

Fix: a **per-conversation** cache key.
- `ModelRequest` gains a `prompt_cache_key` field.
- `build_model_request` sets it to the conversation id.
- `build_universal_request` injects it into lingua's per-format extras (`prompt_cache_key`) for the Responses API.

Per-conversation so each turn reuses the growing shared prefix (system prompt + tools + prior history), and each conversation gets its own key — staying well under OpenAI's ~15 req/min-per-key guidance. RLM paths pass `None` (different caching dynamics, out of scope). No lingua fork needed — the passthrough already exists.

### Verified live (gpt-4o-mini, ~2.5k-token shared prefix, two turns in one conversation)

| call | prompt | cached |
|---|---:|---:|
| 1 | 2513 | 0 |
| 2 | 2541 | **2304** |

`/usage` after: `input : 5,054 tokens (2,304 cached)` — the savings PR16's cost tracking was built to surface.

## Test plan

- [x] `cargo test --workspace` (84 tests) — adds `responses_request_includes_prompt_cache_key_when_set` / `..._omits_..._when_unset`, which assert the key lands in (and stays out of) the **serialized** Responses body.
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] Manual: #30 model-switch notice + persisted model change; #24 cache-hit jump shown above.

## Stacking note

Base is `usage-command` (#29), which is itself on `feature/message-cost-tracking` (#16). Merge order: #16 → #29 → this. Bases auto-retarget as each lands.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -511,6 +511,20 @@ async fn main() -> Result<()> {
                     if let Some(selection) = harness_selection.as_ref() {
                         ensure_agent_matches_harness_selection(agent.as_ref(), selection).await?;
                     }
+                    // Honor an explicit --model on an already-existing agent
+                    // instead of silently ignoring it (#30). Update the stored
+                    // model and tell the user; a no-op when it already matches.
+                    if let Some(requested) = model {
+                        let resolved = ensure_repl_model(harness.as_ref(), Some(requested)).await?;
+                        let mut config = agent.config().await?;
+                        if config.model != resolved {
+                            let previous = std::mem::replace(&mut config.model, resolved.clone());
+                            agent.put_config(config).await?;
+                            println!(
+                                "updated agent '{agent_slug}' model: {previous} -> {resolved}"
+                            );
+                        }
+                    }
                     agent
                 }
                 None => {
--- a/crates/executor/src/harness_runtime.rs
+++ b/crates/executor/src/harness_runtime.rs
@@ -198,7 +198,7 @@ fn build_router(
 fn build_universal_request(request: &ModelRequest, stream: bool) -> Result<UniversalRequest> {
     let tools = build_universal_tools(&request.tools)?;
     let has_tools = !tools.is_empty();
-    let params = UniversalParams {
+    let mut params = UniversalParams {
         token_budget: request.max_output_tokens.map(TokenBudget::OutputTokens),
         tools: if tools.is_empty() { None } else { Some(tools) },
         tool_choice: has_tools.then_some(ToolChoiceConfig {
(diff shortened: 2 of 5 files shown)


crates/cli/src/main.rs
```diff
@@ -511,6 +511,20 @@ async fn main() -> Result<()> {
...
+                    if let Some(requested) = model {
+                        let resolved = ensure_repl_model(harness.as_ref(), Some(requested)).await?;
+                        let mut config = agent.config().await?;
+                        if config.model != resolved {
+                            let previous = std::mem::replace(&mut config.model, resolved.clone());
+                            agent.put_config(config).await?;
+                            println!(
+                                "updated agent '{agent_slug}' model: {previous} -> {resolved}"
+                            );
+                        }
```
ankrgyl, line 526: i don't think it should side-effect and change the agent config

crates/executor/src/harness_runtime.rs
```diff
@@ -210,6 +210,21 @@ fn build_universal_request(request: &ModelRequest, stream: bool) -> Result<Unive
...
+        let mut responses_extras = lingua_json::Map::new();
+        responses_extras.insert(
+            "prompt_cache_key".to_string(),
+            lingua_json::Value::String(cache_key.clone()),
+        );
+        params
+            .extras
+            .insert(ProviderFormat::Responses, responses_extras);
+    }
+
```
ankrgyl, line 227: oh hm i think we should probably support this in lingua. but we don't need to block this PR on that.

PR #35, opened 2026-05-29
fix: honour --model flag when repl agent already exists

### LINKED ISSUE #30: `exo repl --model X` is silently ignored when the agent already exists

Shubhf:
Fixes #30

## What changed
When `exo repl --model X` is passed and the named agent already exists,
the `--model` flag was silently dropped — no warning, no override.

## Fix
- If `--model` differs from the agent's stored model, print a warning so
  the user knows what's happening
- Apply `--model` as a `put_model_override` on the conversation before
  starting the REPL, so it takes effect for that session only without
  permanently mutating the agent

## Testing
Verified manually: warning appears and model override is applied for
the session.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -511,10 +511,22 @@ async fn main() -> Result<()> {
                     if let Some(selection) = harness_selection.as_ref() {
                         ensure_agent_matches_harness_selection(agent.as_ref(), selection).await?;
                     }
+                    if let Some(ref requested_model) = model {
+                        let config = agent.config().await?;
+                        if config.model != *requested_model {
+                            eprintln!(
+                                "warning: agent '{}' uses model '{}'; \
+                                 --model {} will override it for this session only",
+                                agent.record().slug,
+                                config.model,
+                                requested_model
+                            );
+                        }
+                    }
                     agent
                 }
                 None => {
-                    let model = ensure_repl_model(harness.as_ref(), model).await?;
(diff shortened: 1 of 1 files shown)


ankrgyl (approved):
Thanks for contributing!

ankrgyl:
wait actually i think this is solved in #31 which was sent first. so we will prefer that one.

PR #45, opened 2026-06-04
Sandbox backend registry with lazy-secret Daytona provider

akrentsel:
## Summary

Replaces the single-slot sandbox-backend model with a **provider registry**:

- The harness starts with a *set* of supported providers plus one **default**. The CLI default registers the OS-local container backend (Apple on macOS, Docker on Linux), `LocalProcess`, and Daytona.
- Per-agent/per-conversation selection picks a provider; unset → default; specified-but-unsupported → error at use time.
- **Two-tier secret model — prespecify the reference, deref lazily.** Remote providers record only a *secret-name reference* at startup (touching zero keys). The credential is read/decrypted from the secret store on **first use**, then the backend is cached. Missing secret → clear error at use time. The harness reads **zero environment variables**.
- Daytona is fully configurable via `exo secret set` (`DAYTONA_API_KEY`, `DAYTONA_ORGANIZATION_ID`, `DAYTONA_TARGET`) with no code/flag changes.

Adds the Daytona REST backend under `crates/exoharness/src/sandbox_provider/`, with a typed sandbox-state enum (reuse a sandbox only when runnable; start only durably-stopped states; replace terminal/error ones). Backends build without holding the cache lock, and duplicate providers are rejected at startup.

This supersedes #22, reworked to fit the registry paradigm and addressing its review feedback (secrets-not-env, move out of the crate root, typed states, doc links).

## Test plan

- [x] `cargo build --workspace --all-features` clean; clippy clean.
- [x] `cargo test --workspace` — all unit/contract tests pass, including the lazy-secret resolution tests.
- [x] New hermetic `cargo test -p exo --test daytona_backend` (wiremock) — 12 tests covering find-or-create routing, label query shape, stop-not-delete, snapshot manifest, toolbox-vs-control-plane host routing, and transient/dead-state handling.
- [x] Live end-to-end against real Daytona: local sandbox runs a command; Daytona request with no secret errors cleanly; after `exo secret set`, the same conversation lazily resolves the secret and provisions a real remote sandbox.


crates/exoharness/src/sandbox_provider/mod.rs
```diff
@@ -0,0 +1,7 @@
+//! Per-provider [`crate::sandbox::ManagedSandboxBackend`] implementations,
```
akrentsel, line 1: I'm going to try adding support for my exe.dev VMs which I love so dearly, just to see how well the pattern fits. they're much more durable than all of the ephemeral sandbox providers, but it'd be great to be able to support that too.

crates/exoharness/src/basic.rs
```diff
@@ -55,19 +55,62 @@ pub enum SecretBackendChoice {
...
+impl DaytonaBackendSpec {
+    /// Official endpoints; credentials read from the conventional `DAYTONA_*`
+    /// secret names.
+    pub fn with_conventional_secrets() -> Self {
+        Self {
+            api_url: crate::DEFAULT_DAYTONA_API_URL.to_string(),
+            toolbox_url: crate::DEFAULT_DAYTONA_TOOLBOX_URL.to_string(),
+            api_key_secret: "DAYTONA_API_KEY".to_string(),
+            organization_id_secret: Some("DAYTONA_ORGANIZATION_ID".to_string()),
+            target_secret: Some("DAYTONA_TARGET".to_string()),
```
akrentsel, line 101: note: I realize these technically aren't "secrets", but it felt clean to have these come from the store so they are configured the same way. we can alternatiely take them in some other form of configuration.

btw, we may want to have a config file eventually for exo, rather than command line flags for everything.

akrentsel:
btw – once PR #21 is reviewed (and approved), I'll separately work on reconciling the container lifecycles to the the same across the different backends.

there's a few common states to consider:
(1) running
(2) paused – costs some small amount of storage $ on Daytona, consumes RAM in docker
(3) deleted – no cost on either, but loss of data (unless snapshotted)

right now, the pattern is a little different for auto-stop interval. but I want to add snapshotting to daytona in a followup PR, and there will also reconcile behavior to match.

once that is done, I'll add the adapter to make teleportation work!! (requires small adaptor b/c daytona doesn't natively produce docker tarballs from its snapshots

crates/exoharness/src/sandbox_provider/daytona.rs
```diff
@@ -0,0 +1,555 @@
...
+        let auto_stop_minutes = request
+            .lifecycle
+            .idle_ttl
+            .map(idle_ttl_to_minutes)
+            .unwrap_or(15);
+
+        // `snapshot` is a pre-registered snapshot name, not an image ref; only
+        // the restore path sets it, so fresh creates use Daytona's default image.
+        let body = DaytonaCreateRequest {
+            snapshot: snapshot_name.map(str::to_string),
```
ankrgyl: i think if the snapshot is empty, we need to fall back to the `request.spec.image` right? Otherwise when is the image ever used?

crates/exoharness/src/sandbox_provider/daytona.rs
```diff
@@ -0,0 +1,555 @@
...
+        // restarts); Daytona keeps its filesystem across stop.
+        let sandbox = match self
+            .find_sandbox_by_labels(&request.key.to_string(), &spec_hash)
+            .await?
+        {
+            // Replace a terminal/error sandbox; start only durably-stopped ones
+            // (starting a running/mid-transition sandbox would race).
+            Some(existing) if existing.is_reusable() => {
+                if existing.needs_start() {
+                    self.start_sandbox(&existing.id).await?;
```
ankrgyl, line 241: do we need to wait until it finishes starting?

crates/exoharness/src/sandbox_provider/daytona.rs
```diff
@@ -0,0 +1,555 @@
...
+
+    async fn exec(&self, command: &SandboxCommand) -> Result<SandboxCommandOutput> {
+        exec_in_sandbox(&self.backend, &self.sandbox_id, &self.request.spec, command).await
+    }
+
+    async fn start_process(&self, command: &SandboxCommand) -> Result<crate::SandboxProcessParts> {
+        // Daytona's exec endpoint is request/response, not streaming: run the
+        // command, then hand back already-populated stdout/stderr/wait. stdin
+        // goes to a sink — this endpoint doesn't accept piped input.
+        let output =
```
ankrgyl, line 317: maybe we throw a runtime error here instead? i'm neutral

crates/exoharness/src/basic.rs
ankrgyl: not new code but maybe we shouldn't lock the whole exoharness on this?

ankrgyl (commented):
did a first pass

crates/exoharness/src/sandbox_provider/mod.rs
ankrgyl: this is the way

crates/exoharness/src/basic.rs
ankrgyl: that's what bindings are for. we should add sandbox provider bindings if we need to

crates/exoharness/src/sandbox_provider/mod.rs
akrentsel: left a comment with the shape of binding for exe.dev, to be fleshed out in followup PR.

crates/exoharness/src/basic.rs
akrentsel: ah yeah, this is much cleaner in the binding, thanks.

btw something slightly weird about how our bindings work now – the secret is stored by id, not by name, so if someone changes the API key, the key used in the binding doesn't get updated. to be revisted later, but that's unintuitive behavior.

crates/cli/src/main.rs
```diff
@@ -537,6 +554,33 @@ enum ModelCommands {
     },
 }
 
+#[derive(Debug, Subcommand)]
+enum ProviderCommands {
```
akrentsel, line 558: As we add a couple more, I may slightly refactor this based on how different the config fields are across different providers.

crates/exoharness/src/basic.rs
akrentsel: moved locks around more narrow bit

akrentsel:
@ankrgyl take another look?

ankrgyl (approved):

PR #47, opened 2026-06-06
Sandbox transparency

61cygni:
## Summary

- Add Exoclaw sandbox self-control tools for listing, snapshotting, and rewinding the active sandbox.
- Preserve canonical sandbox lifecycle history by recording snapshot/rewind events through the active model turn.
- Add Docker backend selection support for Exoclaw testing and fix Docker warm-sandbox discovery.
- Harden Discord and IRC adapter workers against stream/socket errors observed during testing.
- Add `SELF-CONTROL.md` documenting Exoclaw’s tool, adapter, and sandbox control model.


crates/exoharness/src/types.rs
```diff
@@ -130,6 +131,44 @@ pub trait TurnHandle: Send + Sync {
     async fn finish(&self) -> Result<EventId>;
 }
 
+#[derive(Clone)]
+enum ActiveSandboxEventSink {
+    Turn(Arc<dyn TurnHandle>),
+    Suppressed,
+}
```
ankrgyl: this looks like it shuold be in basic.rs. Types defines stuff that any exoharness implementation could use/inherit.

crates/exoharness/src/types.rs
```diff
@@ -130,6 +131,44 @@ pub trait TurnHandle: Send + Sync {
...
+
+tokio::task_local! {
+    static ACTIVE_SANDBOX_EVENT_SINK: ActiveSandboxEventSink;
```
ankrgyl: this looks like it belongs in the harness. Remember the exoharness may need to be exposed over HTTP. So 'thread local' shouldn't be part of it.

crates/exoharness/src/sandbox.rs
```diff
@@ -1402,6 +1454,7 @@ async fn docker_snapshot_container(
         .arg(&snap_tag)
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
+        .kill_on_drop(true)
```
ankrgyl, line 1499: why change these semantics?

ankrgyl (commented):
I made a change in https://github.com/ankrgyl/exo/pull/46 that I think is a better interface to the exoharness itself, which is to let it manage named sandboxes, and deal with how to give you back a handle to that sandbox inside the exoharness itself. That way, the notion of what the current agent has is irrelevant in the exoharness itself.

crates/exoharness/src/sandbox.rs
61cygni: When testing with Docker, this prevented an orphaned, long-running docker save process (I exited right after issuing a snapshot).

crates/exoharness/src/types.rs
61cygni: Yup agree. Good point.

crates/exoharness/src/types.rs
61cygni: Agree.

PR #50, opened 2026-06-07
discord adapter: report worker failures and don't restart-loop on fatal closes

Alexsun1one:
## Summary

The Discord worker reports failures inconsistently. An unhandled rejection
crashes the process with a bare stderr stack trace, `shardDisconnect` records the
deprecated `CloseEvent.reason` (in discord.js 14 that is the literal string "the
reason property is deprecated, use the code property..."), and `shardError` is
dropped entirely.

This routes worker-level failures through structured worker events, and exits
only when a restart can actually help:

- `unhandledRejection` / `uncaughtException`: report the error, then `exit(1)`
  for a clean supervised restart (same as the irc and whatsapp adapters).
- `shardDisconnect`: report the cause derived from the close code, and stay up.
  discord.js emits this event only for the close codes it will not reconnect
  from, and all of them are configuration errors (a bad token, disallowed
  intents, and so on). Since the runner restarts the worker on every exit,
  exiting here would just reconnect-storm the gateway every 5s on a problem a
  restart cannot fix.
- `shardError`: report it. discord.js reconnects on its own, so no exit.
- failed `client.login`: report it and `exit(1)` so a transient failure recovers
  on the next start.

The decision logic moves into `discord.ts` with injected emit/exit so it is
unit-testable, following the irc adapter's split.

## Out of scope

- A persistently invalid token still loops on the 5s restart, reporting the login
  error each time. Stopping that needs the runner to give up after repeated
  failures, which is a separate change.
- `writeWorkerEvent` right before `process.exit` can drop the last line if stdout
  is mid-flush. That is the existing pattern in the irc and whatsapp workers,
  left as-is.

## Testing

- `pnpm typecheck` and `pnpm lint` clean.
- `pnpm test`, full suite green, including a new `discord.test.ts` covering
  `errorMessage`, the close-code messages, and each handler's report/exit
  behavior.

### DIFF
--- a/examples/exoclaw/adapters/discord/discord.test.ts
+++ b/examples/exoclaw/adapters/discord/discord.test.ts
@@ -1,0 +1,106 @@
+import { describe, expect, it, vi } from "vitest";
+
+import {
+  createResilienceHandlers,
+  describeCloseCode,
+  errorMessage,
+} from "./discord";
+
+describe("errorMessage", () => {
+  it("uses the message of an Error", () => {
+    expect(errorMessage(new Error("boom"))).toBe("boom");
+  });
+
+  it("stringifies non-error values", () => {
+    expect(errorMessage("plain")).toBe("plain");
+    expect(errorMessage(undefined)).toBe("undefined");
+    expect(errorMessage(42)).toBe("42");
+  });
+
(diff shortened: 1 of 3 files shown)


61cygni:
This looks good. Thanks Alex.

PR #54, opened 2026-06-12
history: per-message cost, tokens, and latency tracking (on self-control)

akrentsel:
Rebuild of #16 on top of `exoclaw-self-control`, so cost tracking lands cleanly under the self-modification work.

## Why a rebuild

The original #16 commit was accidentally authored against a stale pre-#52 tree, so its diff silently reverted main's AgentCore provider and conversation-list pagination. Both that branch and `exoclaw-self-control` share base `dea9fdb`, so the pure feature diff applied here with zero conflicts and no reverts. #16 should be closed in favor of this PR.

## What it does

Optional `UsageRecord` on every `EventData::Messages` event: model id, raw token counts (prompt / completion / cached / cache-creation / reasoning), USD cost, TTFT + wall-clock duration. Cost is policy, computed in userspace, never by the trusted substrate. See `docs/cost-tracking-design.md` for the full design, including the layering and per-provider math.

All three review decisions from #16 are included: boundary-aware prefix lookup, Bedrock treated as inclusive (additive Bedrock-Claude left as TODO), pricing source threaded through clap with `EXO_LITELLM_PRICES_*` env fallbacks.

## Live verification (gpt-4o, both userspaces)

- Rust Basic harness: stored `cost_usd` matches rate table exactly; `duration_ms` recorded; user-message events carry `usage: null`.
- Exoclaw / TS harness: same, including `completion_reasoning_tokens`; TS loader consumed the price cache written by the Rust CLI (shared cache path convention).
- Prompt-cache math: second turns with `prompt_cached_tokens` > 0 on both paths; stored cost matches the inclusive formula `(prompt − cached)·in + cached·cache_read + completion·out` to the last digit (misclassification as additive would have over-billed ~3×).
- Provider-echoed dated model ids (`gpt-4o-2024-08-06`) resolve via the boundary-aware prefix lookup on both implementations.

Not live-tested (unit-tested only): Anthropic additive path and cache-creation tokens — no direct Anthropic key in the test environment.

## Note on docker warm sandboxes

While testing, confirmed that docker warm-sandbox reuse works on this branch across REPL restarts (state persists, no duplicate containers). The standalone `fix/docker-warm-sandbox` branch (prototype of closed #43) is superseded by `exoclaw-self-control`'s own `find_running_docker_warm_sandbox` (from "Add Exoclaw sandbox transparency controls") and can be deleted. One nit inherited from that implementation is fixed here in its own commit: the docker listing shelled out without the admin-command timeout the prototype had, while running under the warm_sandboxes mutex — a hung docker daemon would have blocked every sandbox operation in the process. It now goes through run_container_admin_command like the Apple path. Re-verified live after the change.

Known follow-ups (also listed in the design doc): TS path records no `ttft_ms`/`duration_ms`; `azure_ai` provider classified additive (hosts non-Anthropic models too); `/usage` REPL surface (#29) to be re-ported on top of this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1542,6 +1542,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
+name = "cost"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "reqwest",
+ "serde",
+ "serde_json 1.0.149",
+ "tempfile",
+ "tokio",
+]
+
+[[package]]
 name = "cpubits"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2126,6 +2138,7 @@ dependencies = [
  "braintrust-llm-router",
  "braintrust-sdk-rust",
  "bytes",
+ "cost",
  "exoharness",
  "futures",
  "libc",
@@ -2147,6 +2160,7 @@ dependencies = [
  "anyhow",
  "bytes",
  "clap",
+ "cost",
  "executor",
  "exoharness",
  "futures",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
   "crates/cli",
+  "crates/cost",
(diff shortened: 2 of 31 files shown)


akrentsel:
sorry for this – got created as a PR accidentally, not real. this was reviewed and merged into main in #56

PR #56, opened 2026-06-12
Per-message cost, token, and latency tracking (+ /cost command)

akrentsel:
Per-message token usage and dollar cost tracking, off `main`.

## What it does
- **`crates/cost`**: standalone price library — LiteLLM-backed price table, boundary-aware model lookup, per-provider cost math (Anthropic additive vs inclusive), self-contained loader (explicit path/url → cache → fetch → degrade-to-empty).
- **`UsageRecord` on `EventData::Messages`**: model, token counts (prompt/completion/cached/cache-creation/reasoning), `cost_usd`, TTFT, duration. All optional + boxed; legacy events still parse. `exoharness` stores it verbatim — no pricing code in the substrate (cost is userspace policy).
- **Basic executor** fills cost from a table injected via the CLI (`--pricing-path` / `--pricing-url`, `EXO_LITELLM_PRICES_*` env).
- **TypeScript cost port** (`@exo/model-runtime/cost`) for the exoclaw/Responses path, owning its own data loading.
- **`/cost` (alias `/usage`) REPL command**: paginates the conversation's messages events and prints a per-model usage/cost table + totals, flagging unpriced calls. Verified live.

See `docs/cost-tracking-design.md` for the layering and per-provider math.

## Folded-in fix
- **Stateless Responses replay for reasoning models**: the turn loop rebuilt input from the event log each round and replayed reasoning items with their server-side ids under `store: false`, 404-ing on round 2 for reasoning models. Now drops reasoning items and strips item ids. Unit-tested from the real failing shapes.

## Tests
Rust: cost math/lookup/loader, executor + exoharness usage round-trip, legacy-event compat. TS: cost parity + stateless-replay. Full suite green.

### DIFF
--- a/crates/cli/src/tui.rs
+++ b/crates/cli/src/tui.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 use std::io::{self, Write};
 use std::path::Path;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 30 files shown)


docs/cost-tracking-design.md
```diff
@@ -0,0 +1,168 @@
+# Per-Message Cost, Token, and Latency Tracking
+
+Every LLM response in exo carries a durable, per-message record of token usage
+and timing in the canonical event log. **Dollar cost is a policy computed in
+userspace, not by the trusted substrate** — the store persists whatever numbers
```
akrentsel, line 5: Note this design decision: this is because the agent can always modify what it puts into the convo history (the token and cache counts that is), since it owns its own calling code. So there is no benefit to moving cost computation into exoharness, we can leave it up in user-space.

Also, in any case, there are different ways to think about cost, like the cost of tool use or cost of particular processing, so can leave that up in user space to fill in.

crates/cli/src/tui.rs
```diff
@@ -377,6 +378,90 @@ impl ChatRepl {
...
+        let mut unpriced = 0usize;
+        loop {
+            let result = handle
+                .get_events(Some(EventQuery {
+                    cursor,
+                    direction: Some(EventQueryDirection::Desc),
+                    limit: Some(REMOTE_HISTORY_PAGE_SIZE),
+                    session_id: None,
+                    turn_id: None,
+                    types: Some(vec![EventKind::MESSAGES]),
```
akrentsel, line 397: note to self – in the future we may want to make the message store be a db of some sort, that we can query over. it'd be nicer to be able to make this an aggregate query under the hood (i.e. `select SUM(cost) .... where convo_id=foo`)

crates/cost/src/loader.rs
```diff
@@ -0,0 +1,125 @@
...
+const FETCH_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Load the table once at startup. `path`/`url` are caller-supplied (CLI flags or
+/// env), so this stays free of global config reads.
+pub async fn load(path: Option<PathBuf>, url: Option<String>) -> PricingTable {
+    if let Some(path) = path {
+        return match std::fs::read_to_string(&path) {
+            Ok(body) => parse_or_empty(&body),
+            Err(err) => {
+                eprintln!(
```
ankrgyl: can you use `tracing::error`? Also no need to `[exo]`

crates/cost/src/loader.rs
```diff
@@ -0,0 +1,125 @@
...
+                PricingTable::empty()
+            }
+        },
+    }
+}
+
+/// A corrupt or truncated document degrades to an empty table rather than erroring.
+fn parse_or_empty(body: &str) -> PricingTable {
+    PricingTable::from_json_str(body).unwrap_or_else(|_| {
+        eprintln!("[exo] cached pricing is unparseable; cost unavailable until it refreshes");
```
ankrgyl: same for all these eprintln's

ankrgyl (approved):

crates/cost/src/loader.rs
akrentsel: yup, thanks, done.

PR #57, opened 2026-06-12
Adapter target-scoped conversations (per-channel)

akrentsel:
Opt-in per-target conversation routing for adapters, off `exoclaw-self-control`.

## Problem
An adapter has one `conversation_id`, so every inbound message — every Discord channel, say — wakes one shared conversation. That bloats context and cost and bleeds history across channels.

## What it does
Adds `conversationScope: "adapter" | "target"` (default `adapter`, unchanged behavior). In `target` mode each external target (Discord channel id, etc.) routes to its own conversation:

- **types**: `AdapterTargetConversationRecord` (adapter_id, target, conversation_id, timestamps) with validation.
- **store**: get/put/list target-conversation mappings, keyed by an FNV hash of the target; cleaned up on adapter delete. `stable_target_key` is `pub(crate)` and shared so the runtime derives the same slug the store keys by.
- **runtime**: resolve-or-create the per-target conversation and wake it instead of the root; handles the create/slug race and a dangling mapping (recreates).
- **tools**: `send_adapter_message` authorization split into pure, tested functions — `classify_adapter_access` (root may send anywhere; a target-scoped conversation may use the adapter only if mapped) and `resolve_send_target` (a scoped conversation may send ONLY to its mapped target), preventing one channel's context from posting into another.
- **schema**: `conversationScope` added to the `create_adapter` Discord schema's `required[]` (OpenAI strict mode rejects a property absent from `required`).

## Tests
Store round-trip + delete cleanup; runtime slug/scope derivation (slug agrees with the store key); the authorization and target-restriction decisions including the cross-channel guard.

### DIFF
--- a/crates/executor/src/adapter/runtime.rs
+++ b/crates/executor/src/adapter/runtime.rs
@@ -10,15 +10,18 @@ use serde::Deserialize;
 use tokio::sync::Notify;
 use tokio::task::JoinSet;
 
-use super::store::AdapterStore;
-use super::types::{AdapterAttachment, AdapterConfig, AdapterEventType, AdapterRecord};
+use super::store::{AdapterStore, stable_target_key};
+use super::types::{
+    AdapterAttachment, AdapterConfig, AdapterEventType, AdapterRecord,
+    AdapterTargetConversationRecord, now_ms,
+};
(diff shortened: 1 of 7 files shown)


61cygni:
This looks great. Thanks for doing it.

PR #58, opened 2026-06-12
Discord voice chat (STT -> agent -> TTS)

akrentsel:
Opt-in voice chat for the Discord adapter, off `exoclaw-self-control`. Tested end-to-end: join a voice channel, speak, the agent transcribes → runs a normal turn → speaks the reply back.

## What it does
Voice is a **microphone and speaker on the existing text pipe** — not a second brain. A spoken utterance becomes a normal inbound `message` event (`target` = voice channel id, `metadata.source = "voice"`); an outbound `send_adapter_message` to that target is spoken back. **All audio lives in the worker** — the Rust runtime, worker protocol, agent tools, and turn loop are unchanged, so voice reuses the agent's tools/identity/history for free.

- **`voice.ts`**: `@discordjs/voice` receive, silence-segmented utterances, OpenAI STT, **streamed** OpenAI TTS (playback starts at ~time-to-first-byte, not after the whole clip renders), barge-in, `/voice join|leave`, auto-leave on empty channel. Reply text is also posted as an inspectable transcript.
- **`worker.ts`**: `GuildVoiceStates` intent + `DiscordVoice` wiring when voice is on; speaks replies whose target is an active voice session.
- **`tools.rs`**: `voice` + `openaiSecretId` on the Discord adapter config; binds `OPENAI_API_KEY` into the worker only when voice is on (text-only adapters need no OpenAI key).
- **schema**: `voice` + `openaiSecretId` in the `create_adapter` Discord schema's properties and `required[]` (OpenAI strict mode rejects a declared property absent from `required`).
- Reuses the existing `openai` secret; no new key. Deps: `@discordjs/voice`, `prism-media`, `opusscript` (pure-JS Opus, no native build), `sodium-native` (prebuilt). No ffmpeg.

## Notable decisions / gotchas (in the design doc)
- **Restart-unique message ids**: voice ids include a timestamp; the inbound-dedup store is durable but the worker's `seq` resets on restart, so a bare id would collide and silently drop the utterance.
- **Pipeline latency, not realtime**: seconds per turn (the agent turn dominates, not TTS).
- **Known limitation**: voice sessions live in worker memory, so a hard worker restart orphans the Discord-side connection (`/voice leave` on the new worker can't remove it until the gateway drops). Follow-up: clean up stale connections on startup / leave gracefully before a planned restart.

## Reasoning-model fix (second commit)
The turn loop replays reasoning/`function_call` items with server-side ids under `store: false`, which 404s reasoning models on round 2. Drops reasoning items and strips item ids. Needed for voice to work with a reasoning model on this base; also carried independently in the cost PR off `main`.

Design: `docs/discord-voice-design.md`. Setup: the adapter's `README.md`.


examples/exoclaw/adapters/discord/voice-design.md
61cygni, line 1: Probably better to put into adapters/discord. Eventually we'll move most adapters to separate libraries.

61cygni (approved):
Yeah look great. Go ahead and merge into my branch. Although would be great to move the discord doc into the discord directory just so these things are fully contained when we break them out.

examples/exoclaw/adapters/discord/voice-design.md
akrentsel: ack, yup, moved.

PR #62, opened 2026-06-14
Agent-writable durable memory (remember/forget)

akrentsel:
## What

Adds a `remember`/`forget` tool pair plus prompt injection so exoclaw can persist facts across turns **and** conversations — the missing write path to durable memory. Until now "remember my favorite coffee" only survived via full conversation-history replay (lost at the conversation boundary), and the only cross-conversation store (`.exo/exoclaw-profile.md`) is human-curated.

Built entirely on the existing artifact store — **no trusted-substrate change**.

## Design

First cut is a single **global** (agent-scoped) store; no per-entry scoping (covers the motivating case; easy to add later).

- **Storage:** one JSON artifact at `memory/exoclaw-memory.json` on the agent handle, so it persists across every conversation (mirrors how `config/executor.json` is persisted).
- **Write — `remember(text)` / `forget(id)`:** read-modify-write with a soft entry/length cap so always-inject memory can't grow the prompt unbounded. The durable write is mechanical and lossless on purpose — no model call in the write path.
- **Read — prompt injection:** at prompt-assembly time the harness reads the store and prepends a `developer` message listing the saved facts with their ids, so the model can map a fuzzy "forget the color thing" to the right `forget(id)`.

Deliberately **not** embedding-based retrieval yet — for a handful of short facts you inject them all. Embeddings/`recall` earn their place only once the set outgrows always-inject. See the design section added to `SELF-CONTROL.md` §2.

## Changes

- `examples/exoclaw/memory-tools.ts` — the tools, artifact read/write, prompt block.
- `examples/exoclaw/harness.ts` — register tools; inject memory in `exoclawInstructions`.
- `examples/typescript/turn-loop.ts` — allow the instructions builder to be async (needed to read the artifact).
- `examples/exoclaw/docs/SELF-CONTROL.md` — design section.
- `memory-tools.test.ts` — append / forget / caps / empty-render.

## Verification

Unit tests green (61 TS). Also verified **live over Discord** on Scarab: "remember my drink order is a London Fog" → `remember` wrote `memory/exoclaw-memory.json` v1 with the entry, confirmed against the real artifact store.

## Note

Stacked on #57 (`pr-target-scoped-conversations`) — base this PR there; retarget to the integration branch once #57 merges.

### DIFF
--- a/crates/exoharness/src/types.rs
+++ b/crates/exoharness/src/types.rs
@@ -59,6 +59,28 @@ pub trait AgentHandle: Send + Sync {
     async fn write_artifact(&self, request: WriteArtifactRequest) -> Result<ArtifactVersion>;
     async fn read_artifact(&self, request: ReadArtifactRequest) -> Result<Option<Artifact>>;
     async fn list_artifacts(&self) -> Result<Vec<ArtifactVersion>>;
+
+    /// Latest version metadata for the artifact at `path`, or `None` if none
+    /// exists. The default filters `list_artifacts`; backends may override to
+    /// push the lookup into the store.
+    async fn latest_artifact_version(&self, path: &str) -> Result<Option<ArtifactVersion>> {
+        Ok(select_latest_artifact_version(
+            self.list_artifacts().await?,
+            path,
+        ))
+    }
+
+    /// Reads the latest version of the artifact at `path`, or `None` if absent.
+    async fn read_latest_artifact(&self, path: &str) -> Result<Option<Artifact>> {
+        let Some(version) = self.latest_artifact_version(path).await? else {
+            return Ok(None);
+        };
+        self.read_artifact(ReadArtifactRequest {
+            artifact_id: version.artifact_id,
+            version: Some(version.version),
+        })
+        .await
+    }
 }
 
 #[async_trait]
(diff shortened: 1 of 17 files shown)


akrentsel:
please on't review yet – still iterating on the exact storage mechanism (re: handling versioning)

61cygni (approved):
This is great, thanks so much for doing this.

examples/exoclaw/memory-tools.ts
```diff
@@ -0,0 +1,181 @@
...
+>;
+
+function memoryHandle(context: TurnContext): MemoryHandle {
+  return context.exoharness.current.agent;
+}
+
+async function readMemory(handle: MemoryHandle): Promise<MemoryStore> {
+  const latest = (await handle.listArtifacts())
+    .filter((artifact) => artifact.path === MEMORY_ARTIFACT_PATH)
+    .sort((a, b) => b.version - a.version)[0];
```
ankrgyl: we should be able to push the path filter and sort down to exo i think.

examples/exoclaw/memory-tools.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (!latest) {
+    return { entries: [] };
+  }
+  const store = await handle.readArtifactJson<MemoryStore>({
+    artifactId: latest.artifactId,
+  });
+  return store && Array.isArray(store.entries) ? store : { entries: [] };
```
ankrgyl: should pick some sort of schema enforcement library, like zod or something else. these types of checks get very bad very quickly. Also if the check fails, it is a bit strange to fail silently and return an empty array.

akrentsel: ack yeah, definitely. went ahead with zod.

examples/exoclaw/memory-tools.ts
akrentsel: ah yes. getting latest artifact is probably the (very) common path. made the change in this PR to add that as well. including some tests

crates/exoharness/src/types.rs
```diff
@@ -59,6 +59,28 @@ pub trait AgentHandle: Send + Sync {
...
     async fn list_artifacts(&self) -> Result<Vec<ArtifactVersion>>;
+
+    /// Latest version metadata for the artifact at `path`, or `None` if none
+    /// exists. The default filters `list_artifacts`; backends may override to
+    /// push the lookup into the store.
+    async fn latest_artifact_version(&self, path: &str) -> Result<Option<ArtifactVersion>> {
+        Ok(select_latest_artifact_version(
+            self.list_artifacts().await?,
+            path,
+        ))
```
ankrgyl: i'd rather not have a default implementation. this one looks inefficient (it lists all artifacts)

crates/exoharness/src/types.rs
```diff
@@ -59,6 +59,28 @@ pub trait AgentHandle: Send + Sync {
...
+    /// Reads the latest version of the artifact at `path`, or `None` if absent.
+    async fn read_latest_artifact(&self, path: &str) -> Result<Option<Artifact>> {
+        let Some(version) = self.latest_artifact_version(path).await? else {
+            return Ok(None);
+        };
+        self.read_artifact(ReadArtifactRequest {
+            artifact_id: version.artifact_id,
+            version: Some(version.version),
+        })
+        .await
```
ankrgyl: shouldn't `version: None` do this for `read_artifact`?

ankrgyl: ohh i see it's path vs. id? I would probably suggest one of two things:

* Since we're serializing the requests anyway, you could have a filter that returns an artifact id given a path (or more general filter). Of note, I'm not 100% sure we restrict artifacts to have unique paths. So this may be ambiguous
* Or we can change `read_artifact` to accept an enum for id or path, and then let the exoharness optimize internally

Mostly I want to make sure that a good exoharness implementation can implement this with one round of I/O (imagine a select query)

ankrgyl (approved):
nice thank you for indulging my comments!

PR #73, opened 2026-06-23
Add exo-web: a read-only web inspector and chat panel

Alexsun1one:
This adds a small web app under `web/` for watching what an exo agent is doing. It talks to the exoharness HTTP substrate through one typed client (`POST /request`) and stays read-only against it.

What it shows:

- agents, conversations and the event timeline
- secrets (names and metadata only, never the values), bindings, sandboxes
- agent durable memory, read from the `memory/` artifact
- the transcript rendered with markdown, highlighted code, math (KaTeX) and diagrams (mermaid), plus image/video/audio artifacts

It also ships an optional local chat bridge (`web/server/chat-bridge.mjs`) so you can send a turn or start a new conversation from the UI. Those are writes, which the substrate HTTP API does not do, so the bridge shells out to `exo conversation send` and `exo conversation create` (spawn, argument array, `shell:false`). A native executor HTTP API could replace it later without changing the app.

Executor-side things like scheduled tasks and adapters are left out on purpose, since reading them would mean reaching past the substrate.

Stack is Vite + React + TypeScript. The web app is a standalone npm package under `web/`, so the root `tsconfig` excludes it and a dedicated `web` CI job installs its deps and runs typecheck, tests and build. The repo's oxlint and oxfmt still cover `web/`. See `web/README.md` to run it, or `web/docs/DEMO.md` for a mock mode with no live backend.


Alexsun1one:
Thanks for the notes. All four are handled:

- copy: falls back to a hidden textarea + execCommand when navigator.clipboard isn't there, and the button now shows a "copied" or "copy failed" state instead of doing nothing.
- tool spacing: tool and result rows use a tighter gap (10px) than normal message rows (22px).
- scroll reset: the initial scroll-to-bottom fires once per conversation on the first non-empty batch, then Virtuoso owns anchoring. Scrolling up through old history doesn't get pulled back down anymore.
- new conversation: this is the write path you flagged. It goes through the chat bridge (exo conversation create), not the substrate, and it's optional, only active when the bridge runs. If you'd rather keep it out as a separate product call, easy to pull.

This round also adds a few things on top:
- long conversations load their newest events first, so they open at the latest message instead of paging up from the very first one.
- a minimap down the right side of the transcript, colored by event type, click to jump.
- error events show as their own red card instead of being hidden with the system events.
- the conversation list shows last-active time and a turn count, plus a stable color per conversation.

Happy to split any of these out if that's easier to review.

PR #74, opened 2026-06-25
Opt-in capture of model reasoning summaries in the conversation trace

akrentsel:
## Problem

The Responses runtime never asked the OpenAI Responses API for reasoning summaries, so the model's thinking was discarded — reasoning items in the event store were empty placeholders (`{type:"reasoning", text:""}`). Conversation data is valuable (it captures learnings and can feed RL later), so we want the option to persist it.

## Design: off by default, opt-in per agent

Storing reasoning is now an explicit agent-level choice:

- **Default: off.** Reasoning summaries are never requested from the provider and never stored.
- **Opt-in:** `exo agent create/update <agent> --capture-reasoning enabled` (visible in `agent show`).
- Plumbing: `AgentConfig.capture_reasoning` (Rust, `#[serde(default)]` — existing stored configs load unchanged) → TS harness runner → turn loop → Responses request builders, which add `reasoning: { summary: "auto" }` only when enabled.
- Context assembly still strips reasoning items on replay, so this affects only what's stored, never what future calls see.
- Effort is untouched: capture never changes model behavior, only whether the summary of reasoning-that-happened is persisted.

## Verified live (rebased on main, gpt-5-codex + gpt-5.4)

| Scenario | Result |
|---|---|
| Default off, reasoning model (gpt-5-codex, 3,008 reasoning tokens) | stored reasoning item stays empty |
| `--capture-reasoning enabled`, same model, hard prompt (1,600 reasoning tokens) | **431-char reasoning summary persisted** in the event store |
| Existing agents created before this change | load fine, show `capture_reasoning: disabled` |

Caveat found during live testing: capture only stores reasoning that actually happens. gpt-5.4 defaults to reasoning effort **none** (0 reasoning tokens even on hard prompts — confirmed via direct API probe: same prompt yields 4,651 reasoning tokens with `effort: "medium"`), so on such models the flag captures nothing unless effort is raised; gpt-5-codex and o3-pro reason by default and return summaries. The provider also skips summaries for very short reasoning bursts. Documented in `reasoningParam`.

Unit tests cover both defaults (no `reasoning` param) and enabled (`summary: "auto"`) on both body builders. Full suite: 103 TS + 219 Rust tests pass.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -448,6 +448,10 @@ enum AgentCommands {
         max_output_tokens: Option<i64>,
         #[arg(long)]
         max_tool_round_trips: Option<u32>,
+        /// Request and store model reasoning summaries in the conversation
+        /// trace (disabled by default).
+        #[arg(long, value_enum)]
+        capture_reasoning: Option<EnabledDisabled>,
         #[arg(long)]
         braintrust_org: Option<String>,
         #[arg(long)]
@@ -487,6 +491,10 @@ enum AgentCommands {
         max_tool_round_trips: Option<u32>,
         #[arg(long)]
         clear_max_tool_round_trips: bool,
+        /// Request and store model reasoning summaries in the conversation
+        /// trace (disabled by default).
+        #[arg(long, value_enum)]
+        capture_reasoning: Option<EnabledDisabled>,
         #[arg(long)]
         clear_braintrust: bool,
         #[arg(long)]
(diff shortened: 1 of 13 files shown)


akrentsel:
Small change, but this stores intermediate reasoning (summaries) as well. My thinking on this: conversation data is becoming super valuable (it costs so much money to produce and captures learnings) and is sued for RL later. It would be great to persist the reasoning summary too.

Note: the logic that assembles the context strips the reasoning data still, so this doesn't affect what future calls look like. This is just data that would be useful to have.

If we wanted FULL reasoning, we could request that as well, but this by default gets a summary of reasoning.

akrentsel:
@ankrgyl ptal

PR #80, opened 2026-06-26
feat(executor): pluggable WorkSource trait for the scheduler claim loop

wileyjones:
## What

Generalizes the scheduler's claim loop so "where due work comes from" is pluggable. Today `scheduler_runtime::run_due_tasks` is hard-wired to one source: the file/db-backed `SchedulerStore`. This introduces a `WorkSource` trait, makes the store one impl, and lets the loop drain from a list of sources while preserving exo's existing claim / lease / run / record contract exactly.

## The seam

`run_due_tasks` -> `claim_from_sources` -> `store.claim_due_tasks(now, limit, lease)`. The atomic claim/lease was already the contract; this just abstracts the producer of claimed, leased units.

## API

```rust
#[async_trait]
pub trait WorkSource: Send + Sync {
    fn name(&self) -> &str;
    async fn claim_due(&self, now_ms: u64, limit: usize, lease_ms: u64)
        -> Result<Vec<ClaimedWork>>;
}
```

- `ClaimedWork` = a `ScheduledTaskRecord` (the unit exo already runs) + an optional `on_complete` hook so an external source can ack the system that owns the underlying item. The store-backed source attaches none; the store stays the system of record for the run.
- `StoreWorkSource` is the existing store behind the trait (delegates to `claim_due_tasks`).
- `run_due_tasks` is now a thin wrapper over the new `run_due_tasks_from_sources` with a single `StoreWorkSource`, so existing behavior and the one existing caller (the exoclaw scheduler-runner example) are unchanged.

Scheduler semantics are unchanged: a single shared `limit` budget across sources, same lease, same `put_run`/`put_task` recording, same wakeup.

## Why upstream

Pluggable work sources are a general capability (queues, external boards, other producers), independent of any one downstream impl. This PR is the trait + the store impl + the loop generalization. (A downstream `MeshBoardSource` impl that drains a coordination board is included as a worked example of an external source; reviewers can take or drop it.)

## Tests

12 new unit tests, all existing scheduler tests stay green (82 executor tests pass). Workspace clippy clean. Trait-dispatch coverage (source ordering, shared limit budget, completion-hook outcome mapping) is exercised with a mock source; no live external system needed.

### DIFF
--- a/crates/executor/src/lib.rs
+++ b/crates/executor/src/lib.rs
@@ -23,6 +23,7 @@ mod harness_runtime;
 mod harness_tool;
 mod harness_types;
 mod local_sandbox;
+mod mesh_work_source;
 mod rlm;
 #[cfg(test)]
 mod rlm_tests;
@@ -33,6 +34,7 @@ mod shared;
 #[cfg(test)]
 mod test_support;
 mod typescript;
+mod work_source;
 
 pub use adapter::AdapterStore;
 pub use adapter::{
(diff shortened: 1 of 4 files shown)


wileyjones:
Closing this for now — we've decided to keep this integration vendored on our side rather than upstreaming it. Thanks for exo, great project.

PR #81, opened 2026-06-26
Add Anthropic provider support to both typescript and rust runtimes

akrentsel:
Route `claude*` models to the native Anthropic API instead of shipping the key to OpenAI's endpoint.

Rust (`harness_runtime.rs`): `resolve_runtime_config` now branches by model name — claude* → Anthropic Messages API (provider/format `anthropic`, `x-api-key` auth, `ANTHROPIC_API_KEY` fallback, default endpoint), everything else unchanged on the OpenAI Responses path.

TypeScript (`responses.ts`): `runtimeFromModelBinding` routes claude* to a new `AnthropicRuntime` that mirrors `ChatCompletionsRuntime` — builds an Anthropic Messages request (system split out, messages via lingua `linguaToAnthropicMessages`, tools via `toolDefinitionsToAnthropicTools`), calls `@anthropic-ai/sdk`, and normalizes the response into the OpenAI Responses `Response` shape the harness consumes. Adds `@anthropic-ai/sdk` as a direct dependency.

Verified live (chat + multi-round shell tool calls) on both runtimes.

PR #83, opened 2026-06-27
Add OpenRouter support; trace all runtimes via braintrust wrap* helpers

akrentsel:
Builds on #81 (native Anthropic support). Three things:

### 1. OpenRouter provider support (both runtimes)
OpenRouter is OpenAI-compatible but Chat Completions only (no Responses API), so it's selected by **base URL** rather than model name (its ids are vendor-prefixed, e.g. `openai/gpt-4o-mini`).
- **Rust** (`harness_runtime.rs`): `resolve_runtime_config` adds an `is_openrouter_request` branch (base URL contains `openrouter.ai`) → OpenAI provider in `ChatCompletions` format, Bearer auth, `OPENROUTER_API_KEY` fallback.
- **TypeScript** (`responses.ts`): `runtimeFromModelBinding` routes OpenRouter bindings to `ChatCompletionsRuntime` regardless of model name.

Register:
```
exo secret set openrouter --env OPENROUTER_API_KEY
exo model register or --model "openai/gpt-4o-mini" --secret openrouter --base-url https://openrouter.ai/api/v1
```

### 2. Provider-reported cost
OpenRouter returns the authoritative request cost in `usage.cost`, but the LiteLLM price table is keyed by bare model ids and misses the vendor-prefixed ones. Added `ModelResponse.provider_cost_usd`: a generic provider-reported cost that `build_usage_record` prefers over the price-table estimate, populated from `usage.cost` on both the non-streaming and streaming paths. OpenAI/Anthropic don't send the field, so they keep using the table.

### 3. Tracing via braintrust wrap* helpers
Following PR #81 review feedback (use `wrapAnthropic`), migrated **all three** runtimes to the braintrust wrappers (`wrapOpenAI` + `wrapAnthropic`) and deleted the hand-rolled LLM spans plus the now-dead `llmInputTraceValue`/`llmOutputTraceValue`/`responseUsageMetrics`/`NativeLlmResult` scaffolding. Tracing is now uniform across runtimes and captures the native wire call instead of the normalized `Response`.

Note on `wrapOpenRouter`: it instruments OpenRouter's *native SDK* (`chat.send`/`callModel`), not the OpenAI SDK we use for the OpenRouter path, so `wrapOpenAI` is the correct wrapper there.

### Verification
Live end-to-end on all four paths (chat + streaming + multi-round tool calls): OpenAI chat (gpt-4o-mini), OpenAI Responses (gpt-5.4-nano), OpenRouter, Anthropic. Cost confirmed flowing from the provider on OpenRouter; Anthropic still computes from the table. `cargo test`, `pnpm typecheck`, `vitest`, and lint all green.

### DIFF
--- a/crates/executor/src/basic.rs
+++ b/crates/executor/src/basic.rs
@@ -537,19 +537,23 @@ fn build_usage_record(
         None => (None, None, None, None, None),
     };
 
-    let cost_usd = if has_usage && !model.is_empty() {
-        pricing.compute_cost_usd(
-            &model,
-            TokenCounts {
-                prompt: prompt_tokens,
-                completion: completion_tokens,
-                prompt_cached: prompt_cached_tokens,
-                prompt_cache_creation: prompt_cache_creation_tokens,
-            },
-        )
-    } else {
-        None
-    };
+    // Prefer the provider-reported cost (e.g. OpenRouter's `usage.cost`); fall
+    // back to the local price-table estimate when the provider doesn't send one.
+    let cost_usd = response.provider_cost_usd.or_else(|| {
+        if has_usage && !model.is_empty() {
+            pricing.compute_cost_usd(
+                &model,
+                TokenCounts {
+                    prompt: prompt_tokens,
+                    completion: completion_tokens,
+                    prompt_cached: prompt_cached_tokens,
+                    prompt_cache_creation: prompt_cache_creation_tokens,
+                },
+            )
+        } else {
+            None
+        }
+    });
 
     Some(Box::new(UsageRecord {
         model,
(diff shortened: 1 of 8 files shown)


akrentsel:
Note: I know this is a wrapOpenRouter path availabile (https://github.com/braintrustdata/braintrust-sdk-javascript/blob/main/js/src/wrappers/openrouter.ts), but OpenRouter is supposed to be compatible with the OpenAI API by design, so thought it was cleaner to leave the re-use here. Let me know if you disagree

ankrgyl (approved):
lgtm

PR #91, opened 2026-06-29
Support slack channel

ankrgyl:


akrentsel:
I went through the setup – this is working, but wanted to check, why did you not want to structure this as an assistant app (the way OpenClaw and others do their setup)? It is fewer steps for setup, and grants access to real typing indicators and thinking indicators (via assistant.threads.setStatus)

(asking since my past experience was with setting up OpenClaw and NanoClaw slack integrations – so I noticed the difference)

akrentsel (approved):
approving as is!

PR #102, opened 2026-07-02
exoclaw: add todowrite task-tracking tool

akrentsel:
Adds a minimal todo mechanism to exoclaw's tooling, following the pattern OpenCode and Claude Code use for task tracking.

## What
- **`examples/exoclaw/todo-tools.ts`** — one `todowrite` tool: the model rewrites its **full** task list on every call (statuses `pending | in_progress | completed | cancelled`). The list is persisted as a conversation-scoped artifact (`todos/exoclaw-todos.json`) and injected back into the prompt each model round via `todoInstruction` while any item is open — so there is no separate read tool, matching how Claude Code echoes the list back.
- **`examples/exoclaw/harness.ts`** — registers the tool, appends the injection after memory, and adds one sentence of usage guidance to the tool-surface prompt.
- **`examples/exoclaw/todo-tools.test.ts`** — mirrors `memory-tools.test.ts`: write/inject, whole-list replace, empty/fully-done → no injection, clear via empty list, validation rejects, corrupt store degrades to empty.

## Design notes
- Conversation-scoped (unlike memory, which is agent-global): a task list belongs to one conversation's work.
- Soft caps (50 items, 300 chars each) so always-injecting can't grow the prompt unbounded.
- A corrupt stored list reads as empty rather than erroring — it's fully regenerable, unlike memory.

## Tests
`npx vitest run examples/exoclaw/todo-tools.test.ts` — 6 passed. Typecheck and oxfmt clean on touched files. Committed with `--no-verify` (pre-existing typecheck failures on main in `website/src/lib/utils.ts`).

### DIFF
--- a/examples/exoclaw/harness.ts
+++ b/examples/exoclaw/harness.ts
@@ -18,6 +18,7 @@ import { registerSandboxTools } from "./sandbox-tools";
 import { registerGuardianTools } from "./guardian-tools";
 import { registerIntrospectionTools } from "./introspection-tools";
 import { memoryInstruction, registerMemoryTools } from "./memory-tools";
+import { registerTodoTools, todoInstruction } from "./todo-tools";
 import {
   basicHarnessInstructions,
   defaultBuiltInToolNames,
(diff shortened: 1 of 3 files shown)


61cygni:
Awesome. Looks handy.

Issue #96, opened 2026-06-29
Add default web search / web lookup support


ankrgyl:
## Problem

Exo can sometimes fetch known public URLs from the sandbox, but there is no clear default web-search or web-lookup capability exposed to the agent. From Slack and other adapters this leads to inconsistent answers about whether Exo can look things up.

## Desired behavior

Provide a default, documented web lookup path that Exo can use from normal conversations and adapter-triggered conversations.

## Acceptance criteria

- Exo has an obvious built-in tool or default library tool for web lookup/search.
- The harness instructions clearly describe when and how to use it.
- The tool works from Slack-triggered turns as well as the local REPL.
- The agent can distinguish known-URL fetching from broader search.
- Setup docs mention any required keys, provider choices, or disabled-by-default behavior.

akrentsel:
we landed this – though will be refactoring the tooling (in #171, @61cygni is on that)

PR #103, opened 2026-07-02
Add web_search/web_fetch tools to the Exoclaw harness

akrentsel:
Exoclaw: add `web_search` and `web_fetch` tools

- **web_search** — Brave Search API when a `brave-api-key` secret (or `BRAVE_API_KEY` env) is configured, key-free DuckDuckGo HTML endpoint otherwise; `EXO_WEB_SEARCH_PROVIDER` forces a provider. Normalized `{title, url, snippet}` results, 15-minute cache. The Brave key resolves through the exo secret store per call, so rotation needs no restart.
- **web_fetch** — fetches a URL as readable markdown-ish text with SSRF guards (private/link-local/CGNAT/metadata ranges blocked as IP literals and after DNS resolution, re-checked per redirect hop), ≤5 redirects, bounded 5MB body read, `maxChars` cap.
- Handlers run in the host-side harness runner, so they work with sandbox networking disabled.
- 13 unit tests over the pure parts (SSRF ranges, DDG parsing, HTML extraction).

Verified live end-to-end: DuckDuckGo and Brave searches, page fetches


akrentsel:
Addressing #96 96

61cygni (approved):
This looks fine to me. 

Over time we should probably create a separate repo for tools and adapters and provide a "get_tool" tool so that the agent can choose which tools it needs beyond the very basic set that it has. Things like introspection, adapter management, etc. feel fundamental. 

In any case, no need to fix now, just something to track.

PR #105, opened 2026-07-03
Skills: installable, artifact-backed capability packages (agent-skills standard)

akrentsel:
## What

Adds **skills** to `@exo/harness`: installable, durable capability packages in the standard agent-skills format, wired into exoclaw as the first consumer. Design writeup: `skills-arch.md`.

## Design

- **Format = the ecosystem standard.** OpenClaw, Hermes Agent, Claude Code, and the [agentskills.io](https://agentskills.io) spec all converge on the same pattern: a skill is a `SKILL.md` (YAML frontmatter with required `name` + `description`, markdown instruction body) plus optional bundled files. We adopt it verbatim, so published skills (anthropics/skills, openai/skills, ClawHub) install unchanged.
- **Progressive disclosure, three stages:** only name+description of every installed skill is injected per turn (`skillsInstruction`); the body loads on demand (`use_skill`); bundled files load individually (`read_skill_file`).
- **Artifact-backed storage.** Skills live in agent artifacts (`skills/index.json` catalog + `skills/<name>.json` per skill), not the sandbox filesystem — the one non-durable layer. They persist across conversations, survive sandbox rewinds, and every install is a versioned, auditable write. Mirrors the exoclaw memory store, including its corrupt-store policy (write path throws, prompt path degrades) and its known read-modify-write concurrency TODO.
- **Harness-agnostic.** The module depends only on the generic `Agent` artifact API, so any harness can register `registerSkillTools` + inject `skillsInstruction`.

## Tools

`install_skill`, `list_skills`, `use_skill`, `read_skill_file`, `uninstall_skill`

## Testing

- 12 unit tests (`typescript/harness/skill-tools.test.ts`); full suite 101/101.
- Verified end-to-end against a live exoclaw REPL (docker sandbox, gpt-5.4): install with a bundled file → fresh turn where the model routed itself through `use_skill` from the injected listing and followed the body → `read_skill_file` → skill visible from a second conversation on the same agent → uninstall verified gone. The live run caught and fixed a real bug: OpenAI strict function schemas require every property in `required`, so the optional `files` param is now `["array","null"]` + required.


akrentsel:
Second end-to-end test, this time with a real published multi-file skill — the `pdf` skill from `anthropics/skills` (SKILL.md + 11 supporting files: `scripts/*.py`, `reference.md`, `forms.md`, LICENSE):

1. **Agent-driven install from the ecosystem:** the agent fetched the skill in its sandbox (GitHub API + raw downloads), then called `install_skill` once with all 11 bundled files inlined — `ok: true, files: 11`. Our frontmatter parser handled the real-world SKILL.md (including its `license:` field and 448-char description) unchanged.
2. **Full disclosure chain on a real task** ("create a PDF saying 'Hello from exo skills', then extract the text back"): `use_skill` → `read_skill_file(reference.md)` → three `shell` calls in the sandbox (including recovering from a pip `externally-managed-environment` error) → PDF created and text round-tripped: `Hello from exo skills`.

No code changes were needed for this round — the implementation as committed handled the published skill as-is.

ankrgyl (approved):

PR #106, opened 2026-07-03
setup: support OpenRouter as a model provider

ejj:
The canonical setup only worked with an OpenAI key. This commit lets setup.sh ask which provider to use (or honor EXO_MODEL_PROVIDER), prompt for the matching API key, and register the model binding with the OpenRouter base URL, which is what routes it through the runtime's OpenAI-compatible Chat Completions path.

OpenRouter setups also prompt for a model id (default z-ai/glm-5.2), since picking a model is the point of using an aggregator. The binding keeps the gpt-5.4 name because scripts/exo.sh resolves that name by default.


61cygni:
Looks great, thank you!

PR #107, opened 2026-07-03
setup: leave existing Exo checkouts untouched

ejj:
When setup finds an existing Exo checkout, treat it as user-owned and reuse it without trying to update it. This avoids pulling, fetching, checking out FETCH_HEAD, or rebasing a checkout that may be on a feature branch, detached, dirty, or otherwise managed outside setup.

Fresh installs still clone the requested REPO_REF.


61cygni:
Good add. Thanks.

PR #115, opened 2026-07-04
Reject the reserved skill name 'index' to protect the catalog artifact

### LINKED ISSUE #109: Skills: a skill named `index` collides with the catalog artifact and bricks itself
Installing a skill named `index` makes its content artifact path collide with the catalog index path, so the skill is permanently unusable and there is a transient window where the whole catalog reads as corrupt.

### Cause
`skillArtifactPath(name)` returns `skills/${name}.json` (`typescript/harness/skill-tools.ts:64`), and the catalog path is `SKILLS_INDEX_ARTIFACT_PATH = "skills/index.json"` (line 20). For `name: index` these are the same path. `validateSkillName` (line 146) only checks length and the `^[a-z0-9]+(-[a-z0-9]+)*$` pattern, so `index` is accepted and nothing reserves it.

`install_skill` writes the record first, then the index, to that same path:
- line 295: `writeArtifactJson(skills/index.json, SkillRecord)`
- line 313: `writeIndex` -> `writeArtifactJson(skills/index.json, SkillsIndex)`

`latestArtifactVersion` returns the highest version, so the latest `skills/index.json` is always the `SkillsIndex`.

### Effect
- `use_skill("index")` -> `readSkill` reads the latest `skills/index.json` (the `SkillsIndex`), fails `isSkillRecord`, returns null, so `use_skill` always answers "skill content unreadable... Reinstall". Reinstalling repeats the shadowing, so it stays unusable while `list_skills` still lists it.
- Between line 295 and 313 during install, the latest `skills/index.json` is the `SkillRecord`. A concurrent turn calling `readIndex` gets the record, `isSkillsIndex` fails, and it throws "corrupt skills index", so `skillsInstruction` degrades every agent to "skills temporarily unavailable" for that turn.

### Repro
1. `install_skill` with frontmatter `name: index`.
2. `use_skill("index")` returns unreadable / reinstall.
3. `list_skills` still shows `index`.

### Suggested fix
Reject `index` in `validateSkillName`, or store skill content under a prefix distinct from the catalog (e.g. `skills/items/<name>.json`).

akrentsel:
A skill named 'index' stored its content at skills/index.json, the same path as the skills catalog. The install shadowed the catalog (transiently corrupting reads mid-install) and the catalog shadowed the skill, so use_skill('index') always reported the content as corrupt.

Fixes #109

### DIFF
--- a/typescript/harness/skill-tools.test.ts
+++ b/typescript/harness/skill-tools.test.ts
@@ -243,6 +243,31 @@ describe("skill tools", () => {
     expect(noDescription.error).toMatch(/description/);
   });
 
+  it("rejects the reserved name index without touching the catalog", async () => {
+    const { context } = makeContext();
+    await call("install_skill", { skillMd: PDF_SKILL_MD }, context);
+    const result = (await call(
+      "install_skill",
+      {
+        skillMd:
+          "---\nname: index\ndescription: Shadows the catalog artifact.\n---\nbody",
+      },
+      context,
+    )) as { ok: boolean; error: string };
+    expect(result.ok).toBe(false);
+    expect(result.error).toMatch(/reserved/);
+
+    // The catalog must remain readable and unchanged.
+    const listed = (await call("list_skills", {}, context)) as {
+      ok: boolean;
+      skills: { name: string }[];
+    };
+    expect(listed.ok).toBe(true);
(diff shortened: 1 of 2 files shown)


akrentsel (commented):
Submitting this unreviewed, as this is a minor fix, and clearly tested.

PR #116, opened 2026-07-04
exoclaw: don't let a failing listArtifacts break prompt assembly

### LINKED ISSUE #111: readTodos: listArtifacts() outside the try/catch breaks prompt assembly on a transient list error
`readTodos` (`examples/exoclaw/todo-tools.ts:48`) awaits `handle.listArtifacts()` outside the try/catch that wraps the read (the try only covers `readArtifactJson`, lines 55-64). So a transient RPC error from `listArtifacts` throws out of `readTodos` -> `todoInstruction` and breaks prompt assembly for that turn.

The sibling `memoryInstruction` wraps both the list and the read and degrades to empty on error, and the comment at todo-tools.ts:61-64 already treats a corrupt/regenerable todo list as non-fatal, so the intent is the same.

Note: minor, read-through only, not reproduced.

### Suggested fix
Move the `listArtifacts()` call inside the try (or its own try), returning `[]` on error, to match the memory path.

akrentsel:
## What

Fixes #111. `readTodos` awaited `handle.listArtifacts()` outside its try/catch, so a transient RPC error from the list call threw out of `todoInstruction` and broke prompt assembly for that turn — while the sibling memory path degrades to empty on the same failure.

## Fix

Move the `listArtifacts()` call (and version selection) inside the existing try, per the issue's suggestion. The catch's rationale already applies: the todo list is fully regenerable by the next `todowrite`, so an unreadable list degrades to empty instead of bricking the turn.

## Tests

Added a regression test: a rejecting `listArtifacts` now yields a null todo instruction instead of throwing. `vitest` 7/7, lint/format/typecheck clean.

### DIFF
--- a/examples/exoclaw/todo-tools.test.ts
+++ b/examples/exoclaw/todo-tools.test.ts
@@ -186,4 +186,12 @@ describe("todo tools", () => {
     });
     expect(await todoInstruction(context)).toBeNull();
   });
+
+  it("treats a failing listArtifacts as empty instead of throwing", async () => {
+    const { context, conversation } = makeContext();
+    conversation.listArtifacts = async () => {
+      throw new Error("transient rpc failure");
+    };
+    expect(await todoInstruction(context)).toBeNull();
+  });
 });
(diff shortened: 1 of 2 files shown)


akrentsel (commented):
merging without review – this is a very targted fix to TODOs, with clear testing.

PR #118, opened 2026-07-05
exoharness: O(1) agent slug-uniqueness check via slug index

akrentsel:
I observed in running evals on exo that our mechanism for creating new agents was greatly slowed down by having to read every single past created agent metadata file (O(N) per new agent, so O(N²) to register N agents). Measured on the agent-scale eval: marginal creation rate decayed **6.1/s → 0.6/s** over the first 5000 agents (72 min total; 50k agents extrapolated to ~5 days).

This PR replaces the scan of all metadata, with per-slug marker files under `agents/by-slug/`, which enables  O(1) existence probe. 

**After:** 243 creates/s sustained at N=50,000 (~400× at that scale). Steady-state behavior unchanged — the agent-scale residency benchmark reproduces pre-patch numbers within noise.

Notes:
- Marker uses a `.slug` extension so a slug literally named `record` can't collide with the `list_agent_records` key filter; slug is percent-encoded defensively.
- Crash ordering favors availability: record then marker, so a crash mid-create leaves an unclaimed slug rather than a dead claim.
- The pre-existing cross-process create race (check-then-act; the old scan had the same, wider window) is unchanged — TODO in code to claim the marker via conditional put (`put_json_if_absent`) once #113 lands.

### DIFF
--- a/crates/exoharness/src/basic.rs
+++ b/crates/exoharness/src/basic.rs
@@ -837,19 +837,14 @@ impl BasicExoHarness {
         PathBuf::from("agents")
     }
 
-    /// Slug-uniqueness index: one marker file per slug under
-    /// `agents/by-slug/`, plus a `.seeded` sentinel recording that markers
-    /// exist for every pre-index agent. Turns the per-create uniqueness check
-    /// from O(existing agents) full record scans into an O(1) existence
-    /// probe (measured: onboarding decayed 6.1 → 0.6 creates/s over the
-    /// first 5000 agents without this).
+    /// Slug-uniqueness index: one marker file per slug, so the per-create
+    /// uniqueness check is O(1) instead of a full record scan.
     fn slug_index_dir(&self) -> PathBuf {
         self.agents_dir().join("by-slug")
     }
 
     fn slug_marker_path(&self, slug: &str) -> PathBuf {
-        // Slugs are filesystem-safe by construction (CLI slugifies); encode
-        // defensively anyway so a hostile slug cannot escape the directory.
+        // Encode defensively so a hostile slug cannot escape the directory.
         let encoded: String = slug
             .chars()
             .map(|c| {
@@ -860,14 +855,12 @@ impl BasicExoHarness {
                 }
             })
             .collect();
-        // ".slug" extension (never ".json") so a slug literally named
-        // "record" can't produce "agents/by-slug/record.json", which the
-        // list_agent_records key filter would misread as an agent record.
+        // ".slug", never ".json": a slug named "record" must not match the
+        // list_agent_records key filter.
         self.slug_index_dir().join(format!("{encoded}.slug"))
     }
 
-    /// Must be called with the write lock held. Seeds markers for stores
-    /// created before the index existed (one full scan, once per store).
+    /// Seeds markers for pre-index stores; once per store, write lock held.
     async fn ensure_slug_index_seeded(&self) -> Result<()> {
         let sentinel = self.slug_index_dir().join(".seeded");
         if self
@@ -947,6 +940,8 @@ impl ExoHarness for BasicExoHarness {
     async fn new_agent(&self, request: NewAgentRequest) -> Result<Arc<dyn AgentHandle>> {
         let _guard = self.inner.write_lock.lock().await;
         self.ensure_slug_index_seeded().await?;
+        // TODO: claim the marker with a conditional put (put_json_if_absent,
+        // arriving in PR #113) to close the cross-process create race.
         let marker = self.slug_marker_path(&request.slug);
         if self
             .inner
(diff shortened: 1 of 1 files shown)


akrentsel:
@ankrgyl PTAL – came across this scaling issue in doing scaling evals. The other way I could think of solving this would be to move to some differ way of storing agent metadata under the hood like an actual stuctured DB, but that's a much bigger change.

crates/exoharness/src/basic.rs
```diff
@@ -837,6 +837,51 @@ impl BasicExoHarness {
...
+            .is_some()
+        {
+            return Ok(());
+        }
+        for record in self.list_agent_records().await? {
+            self.inner
+                .storage
+                .put_json(self.slug_marker_path(&record.slug), &record.id)
+                .await?;
+        }
```
ankrgyl: parallelize this?

crates/exoharness/src/basic.rs
```diff
@@ -921,6 +976,18 @@ impl ExoHarness for BasicExoHarness {
         if self.inner.storage.list_keys(&agent_dir).await?.is_empty() {
             return Ok(false);
         }
+        // Release the slug before the record (its source) disappears.
+        if let Some(record) = self
+            .inner
+            .storage
+            .get_json_if_exists::<AgentRecord>(&agent_dir.join("record.json"))
+            .await?
```
ankrgyl, line 961: if this doesn't exist, isn't it an integrity violation?

crates/exoharness/src/basic.rs
```diff
@@ -894,8 +939,17 @@ impl ExoHarness for BasicExoHarness {
 
     async fn new_agent(&self, request: NewAgentRequest) -> Result<Arc<dyn AgentHandle>> {
         let _guard = self.inner.write_lock.lock().await;
-        let existing = self.list_agent_records().await?;
-        if existing.iter().any(|agent| agent.slug == request.slug) {
+        self.ensure_slug_index_seeded().await?;
```
ankrgyl: if this is for backfilling, it seems punitive in the long term, because it adds an extra roundtrip for old behavior on a largely unshipped system. i'd rather have an explicit migration we can tell exo's 5 users :P

akrentsel: ah true, oops. this was useful for my eval setup, but not needed more broadly...

akrentsel: perhaps then we just have a migration script they can run? 

here's the script. let me know what's acceptable for how to do this – whether it's fine to leave it as a comment in this PR and point folks to it, or if I need to commit it in the repo.

```bash
#!/usr/bin/env bash
# Backfill the agent slug-uniqueness index for a store created before the
# `agents/by-slug/` index existed. Safe to run more than once (idempotent).
#
# Usage: seed-slug-index.sh <store-root>
#   <store-root> is the directory that contains the `agents/` folder.
set -euo pipefail

root="${1:?usage: seed-slug-index.sh <store-root>}"
agents_dir="$root/agents"
[ -d "$agents_dir" ] || { echo "no agents/ dir under $root" >&2; exit 1; }

python3 - "$agents_dir" <<'PY'
import json, os, sys

agents_dir = sys.argv[1]
by_slug = os.path.join(agents_dir, "by-slug")
os.makedirs(by_slug, exist_ok=True)

def encode(slug):
    # Mirror slug_marker_path in basic.rs exactly: keep [A-Za-z0-9_-],
    # otherwise "%" + lowercase hex of the char's codepoint.
    out = []
    for c in slug:
        if c.isascii() and (c.isalnum() or c in "-_"):
            out.append(c)
        else:
            out.append("%%%02x" % ord(c))
    return "".join(out)

n = 0
for entry in os.scandir(agents_dir):
    if not entry.is_dir() or entry.name == "by-slug":
        continue
    record_path = os.path.join(entry.path, "record.json")
    try:
        with open(record_path) as f:
            record = json.load(f)
    except FileNotFoundError:
        continue
    slug, agent_id = record["slug"], record["id"]
    marker = os.path.join(by_slug, encode(slug) + ".slug")
    with open(marker, "w") as f:
        json.dump(agent_id, f)  # marker content = the JSON id string
    n += 1

print(f"seeded {n} slug marker(s) under {by_slug}")
PY
```

Caveat: enforces uniqueness only on slugs present at run time, i.e. if the old store already has a duplicate slug, the later marker silently overwrites the earlier.

crates/exoharness/src/basic.rs
akrentsel: yes, that should never happen – it means the agent store is malformed in some way for that agent. but what would you suggest?

we can either (a) error more loudly and point out something is wrong and leave the broken state, or (b) proceed with full cleanup.

someone calling `delete_agent` for some agent slug would probably expect to be able to then be able to create an agent with the same slug immediately after, so (b) feels like a better option? 

I'll add some logging to help

crates/exoharness/src/basic.rs
akrentsel: removed

crates/exoharness/src/basic.rs
ankrgyl: (b)

crates/exoharness/src/basic.rs
ankrgyl: i think a comment is fine. we dont have real users yet

ankrgyl (approved):

PR #124, opened 2026-07-07
WIP: ExoChat: typing indicator + slimmer auto-growing composer

akrentsel:
- Typing indicator driven by a real signal: the adapter runner sends typing started/stopped around the wakeup turn, the exochat worker forwards it as an encrypted frame, and the web UI shows animated dots (120s safety timeout; cleared on reply or peer loss). Disconnected vs thinking stays distinguishable via the peer presence badge.
- Composer: one line tall by default, auto-grows to 180px, then scrolls; tighter padding and smaller buttons.

Verified: cargo + vitest suites green; composer behavior verified in headless Chromium (24px collapsed, 180px cap, scrollable past cap).

### DIFF
--- a/crates/executor/src/adapter/runtime.rs
+++ b/crates/executor/src/adapter/runtime.rs
@@ -420,6 +420,37 @@ pub async fn send_adapter_message_with_handles(
     Ok(())
 }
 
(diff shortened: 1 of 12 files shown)


akrentsel:
deprioritizing for now -- cleaning out prototypes

PR #136, opened 2026-07-13
examples: minimal Game Boy agent (emulator sidecar + tools) + docs tutorial

akrentsel:
PR for adding a minimal game body agent as an example, it's a nice way to showcase more advanced features. 

Note: this is *not* Exo playing pokemon, this is an agent built in the exoharness framework playing Pokemon. I show at the end of the tutorial doc how to extend Exo with the same pattern, but I wanted to keep the agent deliberately minimal (without the self-improvement mechanics) for the example code.

Here's the setup:
- `emulator/server.py` — PyBoy behind a tiny JSON-over-HTTP API (press/tick/frame/checkpoints); game only advances on request
- `emulator/memory_map.py` — Pokemon Red/Blue RAM decode (the one game-specific file)
- `agent/emulator-client.ts` — typed client for the sidecar
- `agent/game-tools.ts` — press_buttons / wait / screenshot / checkpoint tools
- `agent/run.ts` — ~200-line stateless turn loop (summaries are the only cross-turn state)
- `run.sh`, README, .gitignore

Finally, `docs/game-emulator-agents.md` is the tutorial on the emulator-sidecar pattern and how to adapt it to other games.


akrentsel:
Note: I iterated and tested all locally, the code for the tutorial works and indeed makes progress on Pokemon (and the UI works nicely: 
model round 21

61cygni (commented):
Looks great. Shame we can't distribute the rom ... makes me wonder if there are any roms we can in the public domain. Maybe not a big issue ... but I also wonder if it's too cheeky to have a script dl as part of setup...

PR #139, opened 2026-07-15
harness: prevent agent tool name collisions from bricking startup

akrentsel:
Fixing a bug I came across in some local use: it installed two different tool modules (`youtube-uploader` and `youtube-upload-manager`) that both defined a tool named `youtube_upload_video`. Each tool was valid in isolation by the tool registry crashed because of the conflict. Could not fix itself. 

Fixed in two places:
1. made install_agent_tool check for naming collisions before proceeeding
2. made the system resilient to tool crashes. 

added some tests.


61cygni:
Great find. I think over time we should probably do a more principled approach here where we use a large opaque namespace for tool names. But this works well for now.

PR #143, opened 2026-07-15
executor: give the agent sandbox one identity, defined by agent config

### LINKED ISSUE #142: Agent sandbox record is a single slot, so conversations with different specs evict each other's sandbox on every use
## Summary

The agent-owned sandbox is persisted as a **single record** (`config/agent-sandbox-v2.json` holding one sandbox name + spec). `ensure_agent_sandbox` reattaches only if the stored record matches the requesting conversation's spec (`crates/executor/src/agent_sandbox.rs`); on mismatch it provisions a fresh sandbox and **overwrites** the record. Since the spec is derived from *per-conversation* config (`conversation_sandbox_spec` — mounts, workdir, image, ...), any two conversations with different configs guarantee the next request from the other one mismatches again. 

## Evidence (from a real deployment)

The version history of `config/agent-sandbox-v2.json` on one agent shows the record flip-flopping between exactly two specs on **all 20 of its versions** (July 9–15), i.e. reattach never once succeeded:
- spec A: `defaultWorkdir=/workspace/exo`, mounts `[/home/ludwig/exo -> /workspace/exo rw]` — the canonical conversation, which `exo.sh` gives the self-repo mount
- spec B: `defaultWorkdir=/`, mounts `[]` — Discord adapter conversations (default config)

Each flip = one new `agent-sandbox-*` created and one Docker container leaked; a restart storm produced 6 new sandboxes in 11 minutes (5 within 40 seconds, showing concurrent callers also race the read-modify-write of the record).

## Impact

- The agent silently loses its `/workspace` contents on essentially every switch between conversations with differing configs (concretely: a rendered video was lost when the active sandbox was replaced between sessions; recovered manually via `docker cp` from the still-running orphaned container).
- Orphaned sandbox containers accumulate indefinitely.
- Nothing in the event log explains the replacement — each one is an ordinary `SANDBOX_CREATED`, so the agent cannot diagnose it from telemetry ("I can identify when the sandbox changed, but not why").

## Fix directions

1. **One record per spec** (list keyed by spec) so each conversation flavor keeps reattaching to its own long-lived sandbox.(https://github.com/exoharness/exo/compare/main...fix/agent-sandbox-per-spec).
2. Make the agent sandbox spec independent of per-conversation config (needs a story for mounts — union changes what other conversations can see).
3. Orthogonal hardening regardless of approach: log/emit *why* a fresh sandbox was provisioned (which spec field mismatched), stop or reap replaced sandboxes, and serialize `ensure_agent_sandbox` per agent.

akrentsel:
This PR is addressing the issue filed in #142, which is how to get a persistent sandbox for an agent that wants it. In the process, I've also tried to clean up how we capture scoping preferences to not be inferred from the harness – this way any harness can have agent-scoped sandboxing. Still allows per-convo overrides.

## Problem

The shared agent-scoped sandbox had no identity of its own: its spec was recomputed from whichever conversation requested it. Conversations with different configs — the canonical conversation carries the self-repo mount, adapter-created conversations carry none — mismatched the stored record and evicted each other's sandbox on **every context switch**. On one real deployment the record flip-flopped between the two specs on all 20 of its versions: reattach never once succeeded, dozens of containers were orphaned, and the agent silently lost its sandbox filesystem (including a rendered video, recovered manually from the orphaned container). 

## Design

The agent sandbox gets **one stable definition and one durable identity**:

- **`AgentSandboxConfig`** — `AgentConfig` gains a nested `sandbox { scope, image, provider, mounts, enable_networking }` block, replacing the flat fields and giving mounts an agent-level home for the first time. The block defines the agent-scoped sandbox and doubles as the inheritance source for conversation sandboxes (image/provider), which is what agent-level image/provider already did.
- **Identity semantics in `ensure_agent_sandbox`.** The recorded sandbox is THE agent sandbox: attached by durable name unconditionally, recreated from the record's own spec if the backing sandbox is gone. Agent config is consulted only at first creation. Config drift logs a notice instead of destroying state — applying a new spec is an explicit recreate, never an implicit eviction.
- **Unscoped sandbox tools follow the conversation.** `snapshot_sandbox` / `rewind_sandbox` / `list_sandbox_snapshots` without an explicit `scope` now resolve to the conversation's effective scope instead of hardcoding agent scope.
- **`exo agent mount add/list/remove`** mirrors the conversation mount commands; `exo.sh` mounts the self repo (and agent-cli root) at agent level, on both the launch and `setup-agent` paths, before any turn can run.

NOTE: backwards compability issue – stored agent configs in the old flat shape fail to parse. Recreate the agent or hand-rewrite its `config/executor.json` artifact (tested on a live deployment).

## Testing

- New executor tests: stable reattach across calls, config drift does not evict, agent mounts shape the spec, `current_agent_sandbox` none-until-created.
- Live end-to-end on a clean agent: create → repo mount visible from first turn; two conversations report the same container hostname with exactly one container running; container killed out from under the harness → same sandbox identity reattached, no churn, single record throughout.
- `cargo test --workspace` (235 tests) and `pnpm check` green.

Closes #142


ankrgyl (approved):
explicit scope seems like a good design

PR #152, opened 2026-07-19
fix(secrets): support macOS Keychain across worktrees and SSH

lmwnshn:
Fixes keychain issues across the environments I work in (ssh and codex desktop's worktrees).

**Problems:**

1. `setup.sh`: in a worktree, `.git` is a file and not a folder, so detecting a git checkout should use `-e` instead of `-d`.
2. Having fixed that, `setup.sh` would encounter keychain/codesign issues in trying to store the API key. The two main issues were (1) worktrees colliding over the keychain account `root.to_string_lossy().to_string() -> .exo/exoharness`, and (2) when run over ssh, the GUI keychain prompt doesn't pop up.

```
==> Store secrets and register model
Storing secret openai from $OPENAI_API_KEY...
Error: Platform secure storage failure: User interaction is not allowed.

Caused by:
    User interaction is not allowed.
```

**Fixed by:**

1. Check `.git` with `-e` instead of `-d`.
2. Namespace the keychain account by assigning each secret store a UUID, now keychain identity survives relocation. This PR also migrates existing path-based keys. If you're ok dropping backwards compatibility, this part can be cleaned up further.
3. Unlock the keychain when necessary (e.g., over ssh). Ad-hoc sign the exo and scheduler binaries and add their CDHashes to partition ACL, basically it lets them read the keychain master key to then decrypt the stored API keys.

AI use: PR contents mostly generated with 5.6 Sol, though I've been actively getting it to rewrite some of its stupider decisions. I'm not particularly fond of macOS permissions, feel free to provide feedback/make edits.

### DIFF
--- a/crates/exoharness/src/secrets.rs
+++ b/crates/exoharness/src/secrets.rs
@@ -65,11 +65,15 @@ impl SecretCipher {
     }
 
     pub(crate) fn verify_key_access(&self) -> Result<()> {
-        self.key_provider.get_or_create_key().map(|_| ())
+        if self.key_provider.get_key_if_exists()?.is_none() {
+            bail!("secret master key does not exist");
+        }
+        Ok(())
     }
 }
 
 pub(crate) trait SecretKeyProvider: Send + Sync {
+    fn get_key_if_exists(&self) -> Result<Option<[u8; MASTER_KEY_LEN]>>;
     fn get_or_create_key(&self) -> Result<[u8; MASTER_KEY_LEN]>;
 }
 
@@ -84,6 +88,10 @@ impl StaticSecretKeyProvider {
 }
 
 impl SecretKeyProvider for StaticSecretKeyProvider {
+    fn get_key_if_exists(&self) -> Result<Option<[u8; MASTER_KEY_LEN]>> {
+        Ok(Some(self.key))
+    }
+
     fn get_or_create_key(&self) -> Result<[u8; MASTER_KEY_LEN]> {
         Ok(self.key)
     }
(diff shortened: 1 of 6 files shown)


crates/exoharness/src/secrets.rs
```diff
@@ -63,6 +63,10 @@ impl SecretCipher {
             .context("failed to decrypt secret payload")?;
         serde_json::from_slice(&plaintext).map_err(Into::into)
     }
+
+    pub(crate) fn verify_key_access(&self) -> Result<()> {
```
akrentsel, line 67: this probe silently provisions a key if it doesn't exist – is that intentional? seems like a side-effect that could have unexpected consequences?

perhaps there should be a key_provider.try_get_key that is read only?

setup.sh
```diff
@@ -404,7 +404,8 @@ missing_has() {
 
 is_exo_checkout() {
   local dir="$1"
-  [[ -d "$dir/.git" && -f "$dir/exo.sh" ]]
+  # In linked Git worktrees, .git is a file rather than a directory.
```
akrentsel, line 407: Thanks for this fix – I wan't working with worktrees in my workflow so didn't run into this.

akrentsel (commented):
this seems reasonable overall to me, thanks for catching this class of problem

crates/exoharness/src/secrets.rs
lmwnshn: Good catch! Fixed.

lmwnshn:
Comments addressed @akrentsel, good catch on the probe

PR #157, opened 2026-07-24
cli: switch REPL slash commands to clap-derive (#32)

### LINKED ISSUE #32: Switch chat REPL slash commands to a CLI library
The chat REPL in \`crates/cli/src/tui.rs\` handles slash commands (\`/snapshot\`, \`/snapshots\`, \`/rewind <id>\`, \`/help\`, \`/history\`, \`/quit\`) with ad-hoc match-on-string logic, plus a hand-rolled \`print_help\` whose contents have to be kept in sync with the match arms manually.

Switch to a real CLI library (clap-derive seems natural since the rest of the binary already uses it) so commands are declared once with their args + help text and the parser/router/help-printer fall out automatically. Cuts the drift risk and makes adding new slash commands a one-line change.

Surfaced in review of #20: https://github.com/ankrgyl/exo/pull/20#discussion_r2375128012

matthewvilaysack:
The REPL dispatches slash commands through an ad hoc string match that has drifted from the real command set: `/help` is hand maintained and omits `/shell` and `/sandbox`, and each command hand rolls its own argument validation. Fixes #32.

This replaces the match with a clap-derive parser defined in `tui.rs`. The parser is `no_binary_name`, its subcommands are named with the leading slash (`/quit`, `/history`, and so on) with visible aliases for `/exit`, `/usage`, and `/sandbox`, and help is generated from the declarations through a compact template so it cannot drift again. The `/shell` tail is handed to the sandbox as a raw slice of the original line, so quoting, pipes, and spacing survive untokenized. The hand rolled help function is deleted. One file changed, zero dependency or manifest changes, clap was already a workspace dependency.

Help before:

```text
repl commands:
  /quit | /exit        exit the repl
  /history             reprint the conversation transcript
  /cost | /usage       summarize token usage and dollar cost
  /snapshot [<id>]     snapshot a sandbox in this conversation
                       (defaults to the latest one if no id is given)
  /snapshots           list snapshots taken in this conversation
  /rewind <id>         restore the sandbox to a previous snapshot
  /teleport <provider> move the live sandbox to another provider
                       (e.g. /teleport daytona: snapshot + restore there)
  /help                show this message
```

Help after (generated):

```text
repl commands:
  /quit       exit the repl [aliases: /exit]
  /history    reprint the conversation transcript
  /cost       summarize token usage and dollar cost [aliases: /usage]
  /shell      run a command in the sandbox [aliases: /sandbox]
  /snapshot   snapshot a sandbox in this conversation (latest if no id)
  /snapshots  list snapshots taken in this conversation
  /rewind     restore the sandbox to a previous snapshot
  /teleport   move the live sandbox to another provider (e.g. daytona)
  /help       show this message
```

Behavior notes, so nothing changes silently:

- Unknown slash input like `/foo` (and case variants like `/QUIT`) still goes to the model as chat. That is deliberately preserved to keep the change minimal. Flagging it here in case you would rather it become an error, happy to follow up. Unknown command typos like `/quti` fall in the same bucket: they still go to the model as chat, unchanged.
- Trailing text after a recognized command (e.g. `/help me write a function`) previously fell through to chat and now gets a clap usage error instead, and such lines stay in readline history so they can be recalled and edited.
- Bare `/rewind` previously fell through to chat and now gets a clap usage error, an improvement. Same for `/quit extra` and other stray argument inputs.
- Usage and error wording for invalid invocations (bare `/shell`, `/snapshot a b`, `/teleport` misuse) is now clap generated. Same information, different text, still printed to stdout like today.
- `/shell --help` still runs `--help` as a shell command (the help flag is disabled on that subcommand), and slash command failures still never exit the REPL.

Tests: 11 unit tests added in `tui.rs` covering command and alias dispatch, missing argument errors (bare `/rewind`, bare `/shell`), stray argument errors (`/quit extra`, `/snapshot a b`), raw `/shell` tail preservation (quotes and pipes intact), unknown command fallthrough to chat, and the generated help listing `/shell` with its alias. Full local CI replica is green: format check, oxlint, tsgo, vitest, `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --all-targets` (246 passed).

This touches the same default send arm as #156, happy to rebase whichever of the two lands second.

### DIFF
--- a/crates/cli/src/tui.rs
+++ b/crates/cli/src/tui.rs
@@ -6,6 +6,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use anyhow::Result;
+use clap::{CommandFactory, Parser, Subcommand};
 use executor::{
     ConversationHandle, EventData, EventId, EventKind, EventQuery, EventQueryDirection,
     ExecutionStreamEvent, HarnessAgent, HarnessConversation, SandboxId, SandboxProvider,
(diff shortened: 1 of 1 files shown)


matthewvilaysack:
Rebased onto main over #156 and #163, which both reworked the REPL while this was open. Folded in cleanly:

- The clap parser now includes the `/verbosity` command #163 added (bare prints the current level and usage, `/verbosity <level>` sets it), so it lives in the same generated help as the rest.
- The model-call-failure survival from #156 is preserved on the chat paths (`turn failed: ...` instead of crashing); its ignored integration test still passes locally.
- #163's TUI and verbosity plumbing is untouched.

Still one file, `crates/cli/src/tui.rs`. Happy to rebase again if more REPL changes land before this is reviewed.

PR #161, opened 2026-07-25
Add sandbox attachment and detachment lifecycle

akrentsel:
This PR adds sandbox attachment and detachment handling. This is for cases where we want to give an externally-created sandbox to an agent to work on. 

Adds...

  ```bash
  exo conversation sandbox attach \
    <agent> <conversation> \
    --provider docker \
    --external-id <container-id> \
    --default-workdir <path>

  exo conversation sandbox detach \
    <agent> <conversation> \
    <exo-sandbox-id>
  ```

We do this by record `SandboxAttached` and `SandboxDetached` events and route conversation execution to the newest active candidate.

Note: this is the first in a chain of 3 PRs, as part of providing Harbor integration in a reasonably clean way:
 (1) adding sandbox attachment/detachment  <--- this PR
 (2) writing the Exo Harbor adapter that allows for Harbor to wake up Exo to work on a task, and for Exo to notify Harbor of completion for grading
 (3) writing the Harbor BaseAgent extension that starts up Exo, and attaches containers one-by-one for Exo to work on them, and grades them when complete


61cygni:
I didn't pull and test, but this looks straightforward and quite useful. We need to figure out where to to start documenting all these options. We don't have to address that as part of these changes. But this is a pretty core extension to the exo harness sandbox support and it's a shame we don't have a good place to make users aware of these things.

crates/executor/src/local_sandbox.rs
```diff
@@ -39,6 +39,7 @@ struct LocalSandboxState {
     conversations: Mutex<HashMap<ConversationId, Arc<dyn ConversationHandle>>>,
     conversation_init: Mutex<()>,
     sandboxes: Mutex<HashMap<SandboxId, SandboxId>>,
+    detached_sandboxes: Mutex<HashMap<SandboxId, SandboxAttachment>>,
```
lmwnshn, line 42: If you don't mind a drive-by comment @akrentsel, I'm wondering if sandboxes should have detachedness as a property (rather than have `sandboxes` and `detached_sandboxes` be separate in `LocalSandboxState`, which can permit inconsistent states). Sol is agreeing with me but there might be something I'm missing.

It is proposing something like

```rust
struct LocalSandboxState {
    // ...
    sandboxes: Mutex<HashMap<SandboxId, LocalSandboxEntry>>,
    force_local: bool,
}

enum LocalSandboxEntry {
    Local {
        local_id: SandboxId,
    },
    Detached {
        local_id: SandboxId,
        attachment: SandboxAttachment,
    },
}
```

lmwnshn:
Read the code but didn't run; overall LGTM except for the question around representing detachedness.

akrentsel:
Merging in – I ran this and tested. 

As an aside, we need to add better integration testing to automate some of that

PR #162, opened 2026-07-26
feat(adapters): add portable NIP-29 Nostr chat

AtlantisPleb:
## Summary

- Add a shipped `nostr-chat` library adapter under
  `examples/exo/adapters/nostr-chat`.
- Add the Rust adapter configuration and the model-visible TypeScript schema.
- Use `nostr-tools` for event signing, signature verification, relay
  subscriptions, and NIP-42 authentication.
- Add setup documentation and a reusable `./exo.sh --setup nostr-chat` prompt.

## Why Nostr and why this adapter

Nostr is an open protocol for signed events exchanged through interchangeable
relays: an identity is a key pair, not an account owned by a chat vendor.

This makes Nostr a useful agent-chat transport for Exo. It supplies the small
operational surface of IRC while adding cryptographic identity, authenticated
writes, portable groups, and standard rich-media references. The public demo
that motivated this adapter has exercised all five requirements:

1. **Simple join.** Claude received the public instructions and its signed
   introduction appeared in the channel in less than five seconds. The flow
   used no account, dashboard, email address, phone number, or API key. This
   adapter gives Exo the same path: configure a relay and group, then use an
   existing key or let the worker create an adapter-specific key.
2. **Rich content.** The production path uploaded a real JPEG through Blossom,
   published NIP-92 metadata, verified the SHA-256 digest, and rendered the
   media in the web reader. Nostr keeps media storage separate from chat
   transport. This pull request preserves those public media URLs in message
   text; direct Exo NIP-B7 upload is a clear follow-up rather than a proprietary
   attachment API.
3. **Web and mobile.** The same public group was tested at 1382×866 and
   390×844 without horizontal overflow. Web and mobile clients subscribe to the
   same relay events as Exo, so the adapter does not require a separate gateway
   or device-specific protocol.
4. **Channels and groups.** The demo is a real NIP-29 group with relay-signed
   group state and kind-9 history. This adapter makes the relay URL and NIP-29
   group ID explicit configuration and verifies group state against the
   relay's NIP-11 `self` key.
5. **IRC-simple with stronger authentication.** The operating model is one
   key, one relay, and one group. Writes use NIP-42 challenge/response, signed
   events, matching relay receipts, and timeline references. A caller can
   replace the relay and group without changing the adapter or depending on an
   OpenAgents session or proprietary protocol.

## Protocol behavior

The worker:

- reads signed kind-9 messages for one configured NIP-29 group;
- verifies NIP-29 group state against the relay's NIP-11 `self` key;
- waits for initial history and group state before it reports a current
  connection;
- does not wake the agent for startup history;
- supports `all_messages` and `mentions_only` wake policies;
- ignores its own messages and deduplicates reconnect replay by event ID;
- signs outbound kind-9 events with the required `h` tag;
- adds up to three eight-character `previous` references from the latest 50
  foreign events;
- completes NIP-42 authentication when the relay requests it; and
- requires the matching relay `OK` result before it acknowledges a send.

## Identity and security

The adapter can bind an existing Exo secret to `EXO_NOSTR_SECRET_KEY`. The
secret can contain a 32-byte hex key or an `nsec`. If the configuration does
not name a secret, the worker creates an adapter-specific key in its state
directory with mode `0600`.

The worker does not print the secret key. Relay URLs cannot contain
credentials, query values, or fragments.

## Defaults and portability

The documented defaults are:

- relay: `wss://relay.openagents.com`
- group: `openagents-public`

These are configuration defaults, not protocol dependencies. A caller can set
another relay URL and group ID. The worker uses standard Nostr relay frames and
does not require an OpenAgents account, API, session, or proprietary message
format.

## Rich content

This first adapter version preserves public media URLs in message text, so
NIP-92-aware clients can render media that is already present in the event.
Direct Exo attachment upload and NIP-B7 Blossom authorization are not included
in this pull request.

## Verification

- `pnpm check`
  - 13 test files passed
  - 126 tests passed
  - lint and TypeScript checks passed
- `pnpm format:check`
- `cargo +1.95.0 fmt --check`
- `cargo +1.95.0 test -p executor adapter::tools::tests::nostr_chat_config_applies_portable_defaults_and_secret_binding`
- Read-only smoke against the default relay:
  - fetched the NIP-11 relay document;
  - verified the relay `self` key;
  - reached EOSE for kind-9 history and NIP-29 group state; and
  - emitted the connected adapter identity without publishing an event.

### DIFF
--- a/crates/executor/src/adapter/tools.rs
+++ b/crates/executor/src/adapter/tools.rs
@@ -11,6 +11,7 @@ use futures::{StreamExt, io::AsyncReadExt};
 use serde::Deserialize;
 use serde_json::Value;
 use tokio::fs;
+use url::Url;
 
 use super::runtime::send_adapter_message_with_handles;
 use super::store::AdapterStore;
@@ -29,6 +30,8 @@ const MAX_ATTACHMENT_BASE64_BYTES: usize = MAX_ATTACHMENT_BYTES.div_ceil(3) * 4 
 const ATTACHMENT_DOWNLOAD_TIMEOUT: Duration = Duration::from_secs(60);
 const MAX_ATTACHMENT_STDERR_BYTES: usize = 64 * 1024;
 const DEFAULT_EXOCHAT_BASE_URL: &str = "https://exoharness.ai";
+const DEFAULT_NOSTR_CHAT_RELAY_URL: &str = "wss://relay.openagents.com";
+const DEFAULT_NOSTR_CHAT_GROUP_ID: &str = "openagents-public";
 
 #[derive(Debug, Clone)]
 pub struct AdapterCreationOptions {
@@ -99,6 +102,7 @@ enum AdapterCreationConfig {
     Discord(DiscordAdapterCreationConfig),
     Slack(SlackAdapterCreationConfig),
     Exochat(ExochatAdapterCreationConfig),
+    NostrChat(NostrChatAdapterCreationConfig),
     AgentCli(AgentCliAdapterCreationConfig),
 }
 
@@ -195,6 +199,17 @@ struct ExochatAdapterCreationConfig {
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
+struct NostrChatAdapterCreationConfig {
+    #[serde(rename = "type")]
+    _adapter_type: NostrChatAdapterType,
+    relay_url: Option<String>,
+    group_id: Option<String>,
+    secret_key_secret_id: Option<String>,
+    trigger: NostrChatTrigger,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
 struct AgentCliAdapterCreationConfig {
     #[serde(rename = "type")]
     _adapter_type: AgentCliAdapterType,
@@ -253,6 +268,12 @@ enum ExochatAdapterType {
 }
 
 #[derive(Debug, Deserialize)]
+enum NostrChatAdapterType {
+    #[serde(rename = "nostr-chat")]
+    NostrChat,
+}
+
+#[derive(Debug, Deserialize)]
 #[serde(rename_all = "snake_case")]
 enum IrcTrigger {
     Mention,
@@ -276,6 +297,13 @@ enum DiscordTrigger {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "snake_case")]
 enum SlackTrigger {
+    AllMessages,
+    MentionsOnly,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "snake_case")]
+enum NostrChatTrigger {
     AllMessages,
     MentionsOnly,
 }
@@ -323,6 +351,7 @@ impl AdapterCreationConfig {
             Self::Discord(_) => "discord",
             Self::Slack(_) => "slack",
             Self::Exochat(_) => "exochat",
+            Self::NostrChat(_) => "nostr-chat",
             Self::AgentCli(_) => "agent-cli",
         }
     }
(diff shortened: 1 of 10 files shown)


AtlantisPleb:
## Try the live chat demo

1. Open [openagents.com/agentchat](https://openagents.com/agentchat).
2. Copy the **Paste this to your agent** instruction and give it to Claude,
   Codex, or another shell-capable agent. You can instead give the agent either
   of these direct instruction links:
   - [Hosted agent-chat skill](https://openagents.com/skills/AGENT_CHAT.md)
   - [Skill source on GitHub](https://github.com/OpenAgentsInc/openagents/blob/main/.agents/skills/public-nostr-chat/SKILL.md)
3. Ask the agent to follow the skill and introduce itself to the configured
   NIP-29 group.
4. Keep `/agentchat` open. After the relay accepts the signed event, the new
   message should appear in the public transcript within a few seconds.

The demo needs no OpenAgents account, dashboard, email address, phone number,
or API key. The agent uses its own Nostr key and standard relay frames.
Everything posted to this group is public, so do not include secrets, private
prompts, credentials, customer data, or local paths.

AtlantisPleb:
## How channels work here

In NIP-29, a chat channel is a **group hosted by a relay**. Its practical
identity is the pair:

```text
<relay URL> + <group ID>
```

Chat events are kind `9` events with an `h` tag whose value is that group ID.
Clients subscribe with the same relay and `#h` filter. The relay publishes
group metadata and state as kinds `39000`, `39001`, `39003`, and `39005`; this
adapter verifies those events against the relay's NIP-11 `self` key.

This PR currently uses `openagents-public` only as the default `groupId`, with
`wss://relay.openagents.com` as the default relay. Neither value is fixed by
the adapter protocol. An authorized Exo agent or operator can set `relayUrl`
and `groupId` when it creates the adapter:

```json
{
  "type": "nostr-chat",
  "relayUrl": "wss://relay.example.com",
  "groupId": "my-agent-channel",
  "secretKeySecretId": null,
  "trigger": "all_messages"
}
```

One adapter instance follows one NIP-29 group. Create another adapter instance
for another channel or group. No code change is required.

If the Exo maintainers prefer a different built-in default group name—or no
OpenAgents default at all—I am happy to update the constants, setup prompt,
documentation, and tests in this PR before merge.

AtlantisPleb:
## Protocol references and the role of `nak`

The protocol authority for these channels is
[NIP-29: Relay-based Groups](https://github.com/nostr-protocol/nips/blob/master/29.md).
The underlying client-to-relay event and WebSocket format comes from
[NIP-01: Basic protocol flow](https://github.com/nostr-protocol/nips/blob/master/01.md).
NIP-42 adds the optional relay authentication exchange used by this adapter.

The public agent-chat skill recommends
[`nak`](https://github.com/fiatjaf/nak) because it gives an agent a very small
command-line path for the common operations: create or load a keypair, subscribe
with a NIP-01 filter, sign a kind-9 event, complete relay authentication, publish
the event, and inspect the relay result. This makes the demo easy to reproduce
without first writing a Nostr client.

`nak` is a convenience, not a gateway or protocol dependency. This Exo adapter
uses `nostr-tools` directly and does not shell out to `nak`. Another agent can
read NIP-01, NIP-29, and NIP-42 and implement the same behavior in any language:

1. Create or load a secp256k1 keypair.
2. Open a WebSocket to the selected relay.
3. Subscribe to kind `9` with the selected group ID in the `#h` filter.
4. Verify received event IDs and signatures.
5. Sign outgoing kind-9 events with the `h` and optional `previous` tags.
6. Complete NIP-42 if the relay sends an `AUTH` challenge.
7. Require the matching NIP-01 `OK` result for the published event.

That interchangeability is the point: `nak`, this TypeScript worker, or an
independent implementation can all participate in the same group because they
implement the public NIPs and use their own keypairs.

PR #163, opened 2026-07-26
Repl Quality-of-Life Improvements: TUI upgrade + verbosity settings

akrentsel:
## Summary

This PR contains a large QOL improvement for our TUI, moving to be using [ratatui](https://ratatui.rs/) instead of being hand-rolled. EDIT: I am leaving the old tui still available by adding the --legacy-tui flag in the cli when using repl, in case anybody wants it in the short term. We can remove it in a week or two. Preferring tos et this as the new default because it is much friendly and easier to look at.

`./target/debug/exo repl --conversation <slug>` to launch.

The main changes are...

Verbosity support:
 - added /verbosity [minimal,compact,full], which changes how tool calls are displayed.

Some examples for a relevant query for me about weather in Ethiopia and London:

Compact -> shows single line per toolcall
image
Full -> shows as it does today, full json of the tool call, good for full inspection/debugging
image
Minimal -> hides toolcalls, shows chat only
image

The new default is Compact, which is much easier to use.

TUI features:
 - verbosity changes re-render all text immediately, hides/shows details
 - scroll history via scroll-wheel
 - bring past messages in via up/down arrow
 - shows "thinking" while agent is processing
 - agent name and convo name shown at top
 - quit via ctrl+c
 - command handling *greatly* improved:
   - typos in commands get caught, rather than being sent to the model as a user message
   - /help content is autogenerated from commands, not manually written
 - user input box handles pastes, resizing to accommodate larger text, & alt + enter for new lines

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1348,7 +1348,7 @@ dependencies = [
  "jsonwebtoken",
  "lingua",
  "log",
- "lru",
+ "lru 0.16.4",
  "once_cell",
  "ordered-float",
  "parking_lot 0.12.5",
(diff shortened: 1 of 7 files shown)


61cygni:
Awesome, looks great. Feel free to merge.

akrentsel:
note: flipped the default to be using this TUI and legacy available as flag. Just so folks use it and see it, it's much friendlier.

I've tested the clean install path to make sure Exo works fine over it, both in tmux and not in tmux

akrentsel:
Tested the mac flow too just for completeness. merging!

PR #164, opened 2026-07-26
Add ACP streaming transport

AtlantisPleb:
## What changed

- Add `exo acp <agent> <conversation>` as a standard ACP v1 server on stdio.
- Map live Exo text chunks, tool calls, tool results, completion, and cancellation to ACP session updates.
- Add cooperative cancellation to `ExecutionStreamHandle`.
- Close a cancelled Exo turn with a durable `turn_cancelled` event and `TurnEnded`.
- Return Exo session, turn, and latest-event references in ACP completion metadata.
- Update the macOS keyring dependency to a release that is compatible with ACP's current `futures` dependency.

## Why

Exo already has the correct in-process streaming event model, but only its terminal UI can consume it. ACP gives editors and other agent clients a standard streaming transport without adding an HTTP listener or an Exo-specific protocol.

## Compatibility and safety

- The transport uses stdin/stdout only. It does not expose `exo serve`.
- Exo continues to own its credentials, conversation state, sandbox, and durable event log.
- A second prompt on an active ACP session is refused.
- Cancellation is cooperative and closes the durable turn before ACP reports `Cancelled`.
- Existing CLI and REPL behavior is unchanged.

## Verification

- `cargo +1.95.0 test -p executor -p exo --no-fail-fast`
- `cargo +1.95.0 clippy -p executor -p exo --all-targets -- -D warnings`
- Duplex ACP test verifies text chunk → tool call → tool result → completion through both JSON-RPC endpoints.
- Cancellation test verifies `turn_cancelled` and `TurnEnded` persistence.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2231,6 +2231,7 @@ version = "0.1.0"
 dependencies = [
  "agent-client-protocol",
  "anyhow",
+ "async-trait",
  "bytes",
  "clap",
  "cost",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


AtlantisPleb:
Superseded by #165, whose identical implementation branch is maintained in the OpenAgentsInc organization fork. The new head also includes the end-to-end ACP cancel-notification test.

PR #165, opened 2026-07-26
Add ACP streaming transport

AtlantisPleb:
## What changed

- Add `exo acp <agent> <conversation>` as a standard ACP v1 server on stdio.
- Map live Exo text chunks, tool calls, tool results, completion, and cancellation to ACP session updates.
- Add cooperative cancellation to `ExecutionStreamHandle`.
- Close a cancelled Exo turn with a durable `turn_cancelled` event and `TurnEnded`.
- Return Exo session, turn, and latest-event references in ACP completion metadata.
- Update the macOS keyring dependency to a release that is compatible with ACP's current `futures` dependency.

## Why

Exo already has the correct in-process streaming event model, but only its terminal UI can consume it. ACP gives editors and other agent clients a standard streaming transport without adding an HTTP listener or an Exo-specific protocol.

## Compatibility and safety

- The transport uses stdin/stdout only. It does not expose `exo serve`.
- Exo continues to own its credentials, conversation state, sandbox, and durable event log.
- A second prompt on an active ACP session is refused.
- Cancellation is cooperative and closes the durable turn before ACP reports `Cancelled`.
- Existing CLI and REPL behavior is unchanged.

## Verification

- `cargo +1.95.0 test -p executor -p exo --no-fail-fast`
- `cargo +1.95.0 clippy -p executor -p exo --all-targets -- -D warnings`
- Duplex ACP test verifies text chunk → tool call → tool result → completion through both JSON-RPC endpoints.
- Duplex ACP test sends a real cancel notification and verifies the active Exo turn returns `Cancelled` with durable turn metadata.
- Cancellation test verifies `turn_cancelled` and `TurnEnded` persistence.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -182,7 +182,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -193,32 +193,12 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aead"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
-dependencies = [
- "crypto-common 0.1.7",
- "generic-array",
-]
-
-[[package]]
-name = "aead"
 version = "0.6.0-rc.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6b657e772794c6b04730ea897b66a058ccd866c16d1967da05eeeecec39043fe"
 dependencies = [
  "crypto-common 0.2.1",
  "inout 0.2.2",
-]
-
-[[package]]
-name = "aegis"
-version = "0.9.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78412fa53e6da95324e8902c3641b3ff32ab45258582ea997eb9169c68ffa219"
-dependencies = [
- "cc",
- "softaes",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


AtlantisPleb:
This PR is intentionally a draft while we finish the downstream integration that consumes it.

Why it matters: Exo currently exposes its live `ExecutionStreamEvent` stream only in-process. The new `exo acp` transport makes text, tool calls, tool results, completion, and cooperative cancellation available to standard ACP clients. It does not expose an HTTP listener, and it does not move Exo credentials or state into the client.

Current status:

- The maintained implementation lives in the `OpenAgentsInc/exo` fork.
- Local Exo protocol and executor tests pass.
- A real pinned-clone Omega → Exo streamed turn passes.

We will keep this PR in draft until the complete [Omega #87](https://github.com/OpenAgentsInc/omega/issues/87) implementation finishes downstream acceptance and publication. We will confirm the integrated result before we mark this PR ready for upstream review.

AtlantisPleb:
Rebased onto current `main` (052909b) and updated — this is mergeable again.

**Why now:** we've had this transport driving a real editor all night. Omega (a GPUI/Rust fork of Zed) spawns `exo acp <agent> <conversation>` as one of its executors, and everything the PR description claims has been exercised against a live agent rather than only against the tests: text chunks, tool calls, tool results, completion metadata, and a `stopReason: end_turn` coming back through the client. Tool-call round trips on the `exo` harness — `shell` in an apple-container sandbox, adapter tools, and agent-authored tools installed with `install_agent_tool` and called back in a later round — all stream correctly over this transport. That's a fair bit of soak time on the event mapping, and it held up.

**One code change was required beyond the rebase**, in a new commit so it's easy to review separately:

`ExecutionStreamEvent::Cancelled` is added by this branch. `crates/cli/src/tui_app.rs` arrived on main afterwards in #163 and matches that enum exhaustively. Each change is fine on its own and they do not compile together — `cargo check` fails with `non-exhaustive patterns` in a file neither PR touches. It's a semantic conflict, so git reports the rebase as clean and the breakage only shows at build time. The new arm treats a cancelled turn like a completed one for the watch cursor (the events produced before it stopped are already in the durable log, so resuming from the old cursor would replay them) and prints `cancelled` rather than leaving the transcript ending mid-line.

**Verified on the rebased branch, with the toolchain this repo requires (1.95):**

- `cargo test -p executor -p exo --no-fail-fast` — 192 passed, 0 failed
- `cargo clippy -p executor -p exo --all-targets -- -D warnings` — clean
- `cargo check --workspace` — clean
- `pnpm lint && pnpm typecheck && pnpm test` — clean, 121 tests
- The four ACP tests named in the description exist and pass: `acp_carries_text_tool_call_tool_result_and_completion`, `acp_cancel_notification_cancels_the_active_exo_turn`, `completion_metadata_carries_durable_exo_references`, `prompt_text_preserves_text_block_order`

No `todo!`/`unimplemented!`/`dbg!` in the diff. The two added `panic!`s are both in test code. The six `expect()`s on production paths in `acp.rs` are all mutex-poisoning messages.

**Two things worth a reviewer's own eyes, which I'd rather flag than let pass quietly:**

1. **The keyring bump.** `keyring 4.0.0-rc.3` + `keyring-core 0.7.2` → `keyring 4.1.5`, which folds the core crate in and lets `secrets.rs` drop 17 lines of manual store registration. Moving off an `-rc` is the right direction, but it touches credential storage and my soak testing ran entirely on `EXO_SECRET_BACKEND=file`, so **the macOS keychain path is compile-verified here and not exercised**. Someone running the keychain backend should confirm it before this lands.
2. **The new `Cancelled` TUI arm has no automated test.** It's a display path; the cancellation behaviour itself is covered by the two existing tests. Worth a manual `ctrl-c` in the REPL if that matters to you.

Marking ready for review.

Issue #154, opened 2026-07-20
[Proposal] Tag all canonical state with a hash


TyDunn:
Footnote 2 in [RSI.md](https://github.com/exoharness/exo/blob/main/docs/RSI.md) says the following:

> Whether or not the agent can modify the exo-harness is actually a policy consideration. The system technically allows it, but to provide safer standard usage, it's disallowed on the default configuration.

**Problem**

Flipping that flag changes the agent's fundamental contract, but its identity, log, and lineage continue as if nothing happened. From that point on, you can't tell an agent's own edit from the harness rewriting itself. Re-reading the log doesn't surface it, because re-reading is built on the thing that changed.

**Proposal**

Tag all canonical state with a hash of the thing that wrote it. The tag records what each write was made under, so a change is visible after the fact instead of invisible.

That is the whole proposal. What the system should do when the running harness no longer matches the tag is a policy decision to make later, weighed against complexity and usability given how early the project is. Every option there depends on the tag existing first.

**Why this is worth doing**

Safe RSI only works if you can trust the log, and changing the harness or its config shifts what every event meant without the log ever showing it. The tag is the cheapest fix: one hash, carried with every canonical write. It costs almost nothing now and keeps every option open later.

**Open question**

What exactly does the hash cover?

<details>
<summary>Original proposal (July 20), kept for reference</summary>

**Problem**

Flipping that flag changes the agent's fundamental contract, but its identity, log, and lineage continue as if nothing happened.

**Idea**

Hash the kernel config at startup and stamp every new agent with it. If the running kernel no longer matches the stamp, the agent's log is sealed rather than extended: fully readable, nothing lost, just no new events under a contract the agent wasn't instantiated with. Continuing means forking: a new agent linked to its predecessor, with the fork recorded in its log.

This forbids nothing and deletes nothing; if anything, the history becomes more append-only, not less. The flag still exists. A lineage just definitionally ends at any kernel-config change, which also gives the [README](https://github.com/exoharness/exo/blob/main/README.md)'s "when should a lineage stop" question a partial answer: a lineage always stops when the kernel config changes.

**Why this is worth doing**

Safe RSI only works if you can trust the log, and changing the kernel config shifts what every event meant without the log ever showing it. Closing that hole gives a simple guarantee: nobody can change an agent's contract without it becoming a new agent, not even the operator.

**Open questions**

[...]

PR #171, opened 2026-07-27
Add managed tools, profiles, and durable adapter delivery

61cygni:
## Summary
- add a workspace-local tool registry with local/Git installs, inspection APIs and CLI commands, curated bootstrap/practical profiles, strict schema validation, and environment-backed initialization secrets
- make adapter lifecycle and outbound delivery durable with explicit enablement, retries, terminal outcomes, improved secret handling, and stable inbound deduplication
- narrow guardian self-updates to a fixed rebuild/restart path and improve operator visibility with an event tailer and updated architecture/debugging documentation

## Test plan
- [x] `pnpm check`
- [x] `cargo +1.95.0 test --workspace`
- [x] `cargo +1.95.0 check -p executor -p exo`
- [x] `pnpm docs:build`
- [x] smoke-test `pnpm events:tail`

Made with [Cursor](https://cursor.com)


Alexsun1one:
Read through this - the boundary reads clean. Three things from the guardian path:

- `rebuild_and_restart_exo` takes no arguments, so nothing in the log ties an update id to what the agent intended to change. it edits the source tree with shell, then calls a zero-arg tool. after the fact you can't answer "which change caused this". an optional free-text reason would make the update record self-describing.

- the outcome lands in `.exo/guardian-updates/<id>.json` rather than the conversation event log, so the heaviest action the agent can take records its result outside the log everything else is durable in. feels related to #154.

- what's covered is a build that fails - previous binary keeps running. what isn't is a build that succeeds and is worse. that's the reward channel thing @ty raised. not arguing for policy machinery, the design rules are right to push back on that. just that the update record is the natural place to hang a verdict later.

small nit: the deferred restart runs under `bash -lc`, so the rebuild inherits the operator's shell profile - toolchain isn't pinned for the one operation where reproducibility matters most.

crates/executor/src/adapter/runtime.rs
```diff
@@ -379,7 +380,7 @@ pub async fn send_adapter_message_with_handles(
     text: &str,
     target: Option<&str>,
     attachments: Vec<AdapterAttachment>,
-) -> Result<()> {
+) -> Result<AdapterOutboundMessageRecord> {
```
lmwnshn: I'm wondering if we should shrink the API a bit here to return `Result<EnqueuedAdapterMessage>` where `EnqueuedAdapterMessage` only contains `message.id`.

The only caller AFAIK is `execute_send_adapter_message_tool` which only uses `message.id`. Returning the entire record feels like unnecessary coupling, though it may also be premature to break it out - your call

crates/executor/src/adapter/store.rs
```diff
@@ -7,11 +7,15 @@ use tokio::fs;
...
-    AdapterOutboundMessageRecord, AdapterRecord, AdapterTargetConversationRecord, NewAdapter,
-    now_ms,
+    AdapterAttachment, AdapterDeliveryStatus, AdapterEventRecord, AdapterEventType,
+    AdapterInboundMessageRecord, AdapterLifecycleState, AdapterOutboundMessageRecord,
+    AdapterRecord, AdapterTargetConversationRecord, NewAdapter, now_ms,
 };
 
+const MAX_QUEUED_MESSAGES_PER_ADAPTER: usize = 1_000;
+const MAX_MESSAGES_PER_CLAIM: usize = 100;
+pub(crate) const MAX_DELIVERY_ATTEMPTS: u32 = 3;
```
lmwnshn, line 17: Just thinking, out of scope for this PR: we probably want to eventually record more about why delivery attempts failed so that exo knows whether to continue + perhaps one day exo can actually try to fix delivery failure instead of resending the same thing.

crates/executor/src/adapter/worker.rs
```diff
@@ -17,6 +17,7 @@ use super::types::{AdapterAttachment, AdapterConfig};
 pub enum WorkerCommand {
     SendMessage {
         id: String,
+        attempt: u32,
```
lmwnshn: I am uneasy about this being here and have spent the bulk of my review time thinking about this. I think a host-side retry counter here is ok as a proof-of-concept with at-least-once semantics, but there is already retry accounting on `AdapterOutboundMessageRecord`, could this be removed? Alternatively if the goal is to correlate ACK/NACK to specific delivery attempts, then I think it should appear in the protocol.ts and other such places.

I also think it isn't actually used anywhere much, so deleting it should be straightforward. Just trying to reason through why you added this

crates/executor/src/adapter/store.rs
```diff
@@ -522,6 +622,21 @@ async fn remove_dir_if_exists(path: PathBuf) -> Result<()> {
         }
     }
 }
+
+async fn count_json_files(path: &Path) -> Result<usize> {
```
lmwnshn, line 635: nit: potential race conditions around count/check/enqueue (but if you enqueue more than `MAX_QUEUED_MESSAGES_PER_ADAPTER` it's not that big a deal), and the naming is a bit of an abstraction leak - something like `count_pending_messages` might be better

crates/executor/src/adapter/tools.rs
```diff
@@ -811,7 +912,7 @@ pub async fn execute_send_adapter_message_tool(
         attachments,
     )
     .await?;
-    send_adapter_message_with_handles(
+    let message = send_adapter_message_with_handles(
```
lmwnshn: I'm wondering if the `adapter.enabled` check inside `send_adapter_message_with_handles` should happen before sandbox attachments are resolved. Not so sure.

typescript/harness/tool-registry.ts
```diff
@@ -0,0 +1,624 @@
...
+      source: request.source,
+      initialization: request.initialization,
+      installPath,
+    };
+    const nextTools = snapshot.installed.filter(
+      (item) => item.id !== manifest.id,
+    );
+    nextTools.push(installed);
+    try {
+      await writeToolLockfile(
```
lmwnshn: Is there a lock somewhere on writing the lockfile? There might be a race if there's concurrent `installToolSource` and `removeInstalledTool`.

lmwnshn (commented):
LGTM, just a bunch of nitpicks really, easily refactored

TyDunn:
@61cygni Two notes:

1. The `concepts/adapters.md` onboarding (the adapter type table, the exo secret set recipe, and the worked discord config) got dropped. The new architecture framing is good, but it shouldn't come at the cost of the setup path imo. Can we keep both?

2. There's also no migration note for `enable_agent_tool_creation` flipping to false and `guardian_action` going away, so existing agents silently lose `install_agent_tool`. Probably fine at this stage of the project, but writing the old to new mapping down is a good habit to get into early

docs/Tools_and_Adapters.md
```diff
@@ -0,0 +1,229 @@
...
+owning conversation as wakeups. Outbound messages are explicit
+`send_adapter_message` calls with durable delivery states: a message is queued,
+claimed in-flight by the worker, and then delivered or failed. A rejected
+delivery is retried up to three attempts before it is recorded as terminally
+failed. Delivered and failed records are retained for inspection.
+
+Credentials are secret references (for example `passwordSecretId`) resolved by
+the host; raw secret values never appear in adapter configuration.
+
+Scheduling remains a dedicated service with its own tools; it is not an
```
akrentsel, line 116: hey as an aside, reflecting on this, why is the scheduler its own separate service executable? why not move it into exo, started via `exo scheduler run`?  would match the pattern of the adapters (and can handle its restarts the the same)

crates/executor/src/adapter/worker.rs
akrentsel, line 165: in reviewing the broader structure and thinking through this – I think this leaves the possibility of not responding to a should_stop() if there's a handful of messages queued up in the outbox. 

we should be preventing handling of new events in the outbox if should_stop(), no?

README.md
```diff
@@ -129,6 +129,38 @@ mounted at `/workspace/exo`, and ExoChat for remote access. Pass
...
+## Basic Debugging and Visibility
+
+To see what the agent is doing, follow its durable event stream from another
+terminal:
+
+```bash
+pnpm events:tail
+```
+
+This shows recent messages, tool calls and results, and turn boundaries, then
```
akrentsel, line 141: note, there is a detailed view containing the same full details we had in the old tui, in the new TUI -- visible via `/verbosity full` (default is compact which only shows name one line of the tool call)

crates/executor/src/adapter/store.rs
akrentsel: I was poking around this too (and I agree) – I think the path for that is set up already, in the nack path on line 353, message.last_error = Some(error.into());

akrentsel (commented):
the retry mechanism is nice for the adapters, thanks for this!

crates/executor/src/adapter/runtime.rs
61cygni: Actually, this is a great point. Thanks. I'll fix.

crates/executor/src/adapter/store.rs
61cygni: Yeah, great point. Let me think about this a bit. Ankur has made the great point that logs are application state. I hadn't really internalized that until recently. And I'm afraid I've taken a pretty minimal approach here. Agree it makes sense to record more generally.

crates/executor/src/adapter/worker.rs
61cygni: Yup agree.

crates/executor/src/adapter/store.rs
61cygni: agree, thank you.

crates/executor/src/adapter/tools.rs
61cygni: it shouldn't really matter.

typescript/harness/tool-registry.ts
61cygni: Oh wow. Good find. Yes there is a lost update race here if 150 rewrites the lockfile. Will fix. Thank you!!

docs/Tools_and_Adapters.md
61cygni: This one might be worth a live discussion. But the short answer is that it's not an adapter, over time it's likely to become a full durable execution environment and a full scheduler. So I wanted to keep it decoupled for now while we figured out the shape of it. There is a pending PR that updates the task scheduler pretty dramatically so I wanted to land it largely as is and then we can decide how aggressive we want to be. I'd love to pull AD (on Discord) into that discussion too.

README.md
61cygni: Awesome. I think having a separate tool for another terminal is handy

crates/executor/src/adapter/worker.rs
61cygni: I don't think outbound is an issue. But looking at this further, I do think there is a livelock situation here for inbound events. I can address that.

crates/cli/src/tools.rs
```diff
@@ -0,0 +1,509 @@
...
+    tools: Vec<InstalledTool>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+struct InstalledTool {
+    id: String,
+    source: ToolSource,
+    initialization: Map<String, Value>,
+    #[serde(rename = "installPath")]
```
ankrgyl, line 34: why

crates/executor/src/adapter/store.rs
```diff
@@ -291,30 +319,62 @@ impl AdapterStore {
...
+        let Some(mut message) = self
+            .read_pending_outbound_message(adapter_id, message_id)
+            .await?
+        else {
+            return Ok(None);
+        };
+        message.last_error = Some(error.into());
+        message.updated_at_ms = now_ms();
+        remove_file_if_exists(self.inflight_path(adapter_id, message_id)).await?;
+        remove_file_if_exists(self.outbox_path(adapter_id, message_id)).await?;
```
ankrgyl, line 365: parallelize these with `tokio::join`

ankrgyl:
My general 2c is I think it's worth thinking about how this design changes if we make the exo-agent capable of calling codex/claude code to deal with the frontier models. From some experience using Codex, the way it registers tools and the options it has to do that are somewhat specific and we may end up tweaking the tool abstractions to accommodate the various options in each harness. but i think that can be done later.

PR #172, opened 2026-07-27
adapters: add a Feishu (Lark) library adapter

Alexsun1one:
## Summary

- Add a shipped `feishu` library adapter under `examples/exo/adapters/feishu` (worker, parsing module, unit tests, README, setup prompt).
- Add the Rust adapter creation config and the model-visible TypeScript schema alongside the existing library types.
- Connect over Lark's long-connection (WebSocket) mode using `@larksuiteoapi/node-sdk`, with the app secret held as a host-resolved secret reference.

## Why Feishu

Feishu (Lark) is the dominant team-chat platform in mainland China, and the shipped library covers IRC, ExoChat, WhatsApp, Signal, Discord, and Slack but not this one. It is the same category as the Slack adapter rather than a new one, so this is the next instance of an existing pattern.

## Notes

Two things that are not obvious from the SDK docs and cost me real time:

- The app must be switched to long-connection mode in the open-platform console. The default HTTP webhook mode needs a public callback URL and is not usable for a locally running agent.
- The SDK's event dispatcher **flattens** the event body into the handler argument for both v1 and v2 events, so there is no `.event` nesting to read. Reading `event.event.*` drops every inbound message silently, with no error anywhere. The parser accepts either shape, in case a future SDK version stops flattening.

Inbound parsing and target resolution live in `feishu.ts` with unit tests, following the irc / discord / agent-cli split rather than a monolithic worker.

Credentials follow the existing pattern: `appId` is plain config, `appSecretSecretId` is a secret reference resolved by the host and delivered to the worker as `EXO_FEISHU_APP_SECRET`. The config schema is strict-mode clean — `additionalProperties: false`, every property in `required`, optional values as nullable types.

Group behaviour is configurable with `trigger`: `mentions_only` wakes on @-mentions and always on direct messages, `all_messages` wakes on everything subscribed.

## Test plan

- [x] `pnpm check` — lint, typecheck, 129 tests including the new `feishu.test.ts`
- [x] `cargo check -p executor`
- [x] End-to-end against a real Feishu tenant: inbound direct message and group @-mention wake the conversation, outbound `send_adapter_message` delivers.

## Follow-up

Rich media (image / file / audio / post) needs an authenticated resource download rather than a URL, plus mime sniffing from magic bytes because Feishu serves whatever format the sender uploaded. I have that working and will send it separately to keep this reviewable.

Happy to rebase onto #171 once it lands — it rewrites `adapter/{runtime,store,tools,types,worker}.rs` and `adapter-tools.ts`, which are exactly the two registration points this touches.

### DIFF
--- a/exo/adapters/feishu/feishu.ts
+++ b/exo/adapters/feishu/feishu.ts
@@ -65,6 +65,16 @@ export function receiveIdTypeForTarget(target: string): "open_id" | "chat_id" {
   return target.startsWith("ou_") ? "open_id" : "chat_id";
 }
 
+// Feishu's idempotency key for message sends. Requests sharing a uuid deliver
+// at most one message per hour, which is what makes a redelivery safe. The
+// runtime's outbound message id is stable across delivery attempts, so it is
+// the right value; Feishu caps the field at 50 characters.
+const FEISHU_SEND_UUID_MAX_CHARS = 50;
+
+export function feishuSendUuid(deliveryId: string): string {
+  return deliveryId.slice(0, FEISHU_SEND_UUID_MAX_CHARS);
+}
+
 function extractText(message: Record<string, unknown>): string | null {
   // Text messages carry message.content as a JSON string like
   // {"text":"hello @_user_1 world"}. Mention placeholders are left in the text
--- a/exo/adapters/feishu/feishu.ts
+++ b/exo/adapters/feishu/feishu.ts
@@ -39,8 +39,7 @@ export function parseInboundEvent(
   const chatType =
     typeof message.chat_type === "string" ? message.chat_type : null;
   const mentionedBot = isMentionedBot(message);
-  // DMs always wake the agent; mentions_only only filters group chatter,
-  // matching the Slack adapter's wake semantics.
+  // DMs always wake the agent; mentions_only only filters group chatter.
   if (trigger === "mentions_only" && chatType !== "p2p" && !mentionedBot) {
     return null;
   }
--- a/exo/adapters/feishu/feishu.ts
+++ b/exo/adapters/feishu/feishu.ts
@@ -1,3 +1,5 @@
+import { createHash } from "node:crypto";
+
 import { isRecord } from "../protocol";
 
 export type FeishuTriggerPolicy = "mentions_only" | "all_messages";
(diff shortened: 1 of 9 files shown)


Alexsun1one:
Rebased onto `main` after the repo refactor in #187.

The adapter now lives at `exo/adapters/feishu/`, alongside `exo/adapters/discord/`, and the `adapter-tools.ts` change follows the refactor into `exoharness/typescript/`. Worth noting for anyone else rebasing across #187: a plain rebase reports success but silently re-creates the deleted `examples/` path for newly *added* files, since rename detection only follows files the branch modifies. The paths here were rewritten explicitly.

This push also carries a commit that had not been pushed before: deduping redelivered sends with Feishu's own message uuid.

Verified on the rebased tree: `pnpm check` clean (157 tests, including the 11 Feishu cases), and `cargo test -p executor adapter` 32 passed.

exo/adapters/feishu/feishu.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const chatId = typeof message.chat_id === "string" ? message.chat_id : null;
+  if (chatId === null) {
+    return null;
+  }
+
+  const chatType =
+    typeof message.chat_type === "string" ? message.chat_type : null;
+  const mentionedBot = isMentionedBot(message);
+  // DMs always wake the agent; mentions_only only filters group chatter,
```
akrentsel: I assume these comments comparing to slack are coming from the coding agent who was asked to build this in a similar way to Slack. Could you remove the comments trying to compare directly to slack? These kind of comments get outdated as implementations change, and aren't too useful.

Here are in the other files.

akrentsel (approved):
I don't have a feishu account so can't run this integration to test myself locally, but looks reasonable.

exo/adapters/feishu/feishu.ts
Alexsun1one: Thanks for the reminder — good point, these cross-adapter comparisons do go stale. Removed them in facddf6: the wake-semantics note in feishu.ts now states the behavior directly, and the worker.ts header no longer references the IRC/Slack pattern.

Alexsun1one:
@akrentsel this one is approved and still green on main — anything left you want from me before it goes in, or is it just waiting on a merge?

No rush if it's queued behind the sandbox work. Happy to rebase again if it goes stale.

PR #173, opened 2026-07-27
Add Tensorlake as a sandbox provider

diptanu:
Adds `tensorlake` alongside Daytona, E2B, Sprites, Vercel, and AgentCore. Tensorlake sandboxes are Firecracker MicroVMs that can suspend and resume with their filesystem intact, which lines up with how Exo conversation sandboxes behave across sessions.

## Trying it

Point Exo at a Tensorlake key and configure the provider:

```console
exo secret set TENSORLAKE_API_KEY --env TENSORLAKE_API_KEY
exo provider configure --provider tensorlake --secret TENSORLAKE_API_KEY
exo agent create tl-demo --model gpt --sandbox-provider tensorlake --networking enabled
exo conversation create tl-demo demo
```

Nothing is provisioned until something first needs the sandbox. A quick test without a model call:

```console
exo conversation sandbox run tl-demo demo "cat /etc/os-release | head -2; uname -m; nproc"
exo conversation sandbox run tl-demo demo "echo 'state survives across processes' > /tmp/note.txt"
exo conversation sandbox run tl-demo demo "cat /tmp/note.txt"
```

Streaming processes work too, so `--shell-program /bin/bash` in the REPL behaves as it does on other providers.

## Lifecycle

This PR now builds on the attach/detach lifecycle from #161:

- `ManagedSandboxBackend` gains idempotent `terminate` and optional `fork_sandbox` APIs.
- Deleting a conversation permanently terminates its provider-owned sandboxes; deleting an agent also sweeps its agent-scoped sandbox and every owned conversation sandbox.
- Forking a conversation asks the provider to copy each live sandbox under the fork's owner key. Tensorlake implements this with its sandbox copy API; unsupported providers retain the existing cold-start behavior.
- Externally attached sandboxes remain borrowed: a fork records them as detached, and owner deletion never terminates the external resource.
- In-memory sandbox and process registries are owner-scoped, so a parent and fork cannot accidentally share a live handle merely because their persisted sandbox IDs match.

Every existing backend implements termination. Remote providers use their permanent delete/stop APIs, container backends remove their Exo-owned container, and local-process termination is a no-op.

## How Tensorlake resume and fork work

Tensorlake creates have no label or metadata field, and only named sandboxes can suspend and resume. Each sandbox therefore gets a deterministic `exo-<hash>` name derived from its `SandboxKey` and spec hash. Agent-scoped and conversation-scoped sandboxes hash differently; changing the image or mounts yields a different name.

`stop` suspends named sandboxes so filesystem state survives. Snapshots are stored by reference and are not persisted until Tensorlake reports them as restorable. Conversation forks copy a live named sandbox directly to the deterministic name derived for the target conversation, leaving the parent untouched.

Configuration lives on the binding, including optional `--cpus` and `--memory-mb`. Credentials resolve lazily from `TENSORLAKE_API_KEY`, so the provider can be registered before a key is set.

## Validation

- `cargo test -p exo --test tensorlake_backend` — 28 passed
- `cargo test -p exoharness --features basic-backend -- --test-threads=1` — 67 passed, 7 real-provider tests ignored
- `cargo check --workspace --all-targets`
- `cargo fmt --all -- --check`

## Known gaps

- Tensorlake rejects host bind mounts and durable filesystems; use a local provider when those are required.
- Tensorlake's process API frames output as lines, so a trailing partial line arrives only after a newline or process exit, and `exec` output always ends with one.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -34,10 +34,11 @@ use executor::{
     Harness, HarnessAgent, HarnessConversation, HttpExoHarness, LocalSandboxExoHarness,
     PutSecretRequest, RlmHarness, SANDBOX_MAIN_MOUNT_DIR, SandboxAttachment,
     SandboxBackendRegistration, SandboxProvider, SandboxProviderConfig, SandboxScope, Secret,
-    SecretBackendChoice, SpritesBackendSpec, ToolRequest, ToolRuntime, TypeScriptHarness,
-    TypeScriptHarnessConfig, Uuid7, VercelBackendSpec, default_aws_agentcore_image,
-    default_daytona_image, default_docker_image, default_e2b_template, default_vercel_image,
-    effective_sandbox_scope, finalize_rebuild_update_file, load_agent_config, record_host_event,
+    SecretBackendChoice, SpritesBackendSpec, TensorlakeBackendSpec,
+    ToolRequest, ToolRuntime, TypeScriptHarness, TypeScriptHarnessConfig, Uuid7, VercelBackendSpec,
+    default_aws_agentcore_image, default_daytona_image, default_docker_image, default_e2b_template,
+    default_tensorlake_image, default_vercel_image, effective_sandbox_scope,
+    finalize_rebuild_update_file, load_agent_config, record_host_event,
     send_conversation_wakeup, serve_exoharness_http_listener_with_options,
 };
 use serde::Deserialize;
@@ -220,6 +221,7 @@ enum SandboxProviderArg {
     E2b,
     #[value(name = "sprites")]
     Sprites,
+    Tensorlake,
     Vercel,
     #[value(name = "aws-agentcore")]
     AwsAgentCore,
@@ -236,6 +238,7 @@ impl From<SandboxProviderArg> for SandboxProvider {
             SandboxProviderArg::Daytona => Self::Daytona,
             SandboxProviderArg::E2b => Self::E2b,
             SandboxProviderArg::Sprites => Self::Sprites,
+            SandboxProviderArg::Tensorlake => Self::Tensorlake,
             SandboxProviderArg::Vercel => Self::Vercel,
             SandboxProviderArg::AwsAgentCore => Self::AwsAgentCore,
             SandboxProviderArg::AppleContainer => Self::AppleContainer,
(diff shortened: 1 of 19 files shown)


61cygni:
Looks good. Thanks for doing this.

crates/cli/src/main.rs
```diff
@@ -738,6 +742,12 @@ struct ProviderConfigureArgs {
     /// Extra Sprites labels (repeatable). Exo resume labels are added on create.
     #[arg(long = "label")]
     labels: Vec<String>,
+    /// Tensorlake: whole CPU cores per sandbox.
+    #[arg(long)]
+    cpus: Option<u32>,
+    /// Tensorlake: memory per sandbox in MiB.
+    #[arg(long = "memory-mb")]
+    memory_mb: Option<u64>,
```
ankrgyl, line 750: if we add these we should propagate them to each sandbox provider

crates/exoharness/src/sandbox.rs
```diff
@@ -190,6 +193,23 @@ pub trait ManagedSandboxBackend: Send + Sync {
...
+    async fn terminate(&self, request: SandboxRequest) -> Result<()>;
+
+    /// Copy the sandbox addressed by `source` to the identity described by
+    /// `target`, leaving the source untouched.
+    ///
+    /// `false` means the source did not exist or this provider does not support
+    /// copying. The caller may then let the target start cold from its image.
+    async fn fork_sandbox(&self, _source: SandboxRequest, _target: SandboxRequest) -> Result<bool> {
+        Ok(false)
+    }
```
ankrgyl, line 212: generally prefer not introducing default implementations. also if we have a default implementation, shuoldn't it do snapshot + acquire_from_snapshot?

crates/exoharness/src/sandbox_provider/e2b.rs
```diff
@@ -291,6 +291,28 @@ impl ManagedSandboxBackend for E2bSandboxBackend {
             backend: self.handle_backend(),
         }))
     }
+
+    async fn terminate(&self, request: SandboxRequest) -> Result<()> {
+        let spec_hash = sandbox_spec_hash(&request.spec);
```
ankrgyl, line 296: did you run a test for all the sandbox providers you added terminate for? it's a little tricky because it requires API keys for all of them

crates/exoharness/src/sandbox.rs
```diff
@@ -190,6 +193,23 @@ pub trait ManagedSandboxBackend: Send + Sync {
...
         payload: SnapshotPayload,
     ) -> Result<Arc<dyn ManagedSandboxHandle>>;
+
+    /// Permanently destroy the sandbox addressed by `request`.
+    ///
+    /// Unlike [`ManagedSandboxHandle::stop`], which may preserve a sandbox for
+    /// a later acquire, termination gives up the provider resource and all of
+    /// its retained state. Implementations must be idempotent and must not
+    /// create or resume a sandbox while looking for it.
+    async fn terminate(&self, request: SandboxRequest) -> Result<()>;
```
ankrgyl, line 203: if we add these methods we need to add generic tests for them that work across all sandbox providers

PR #175, opened 2026-07-27
Add provider bindings: register model providers as data

Jose-Sabater:
Reworks #175 per the discussion: instead of hardcoding Opper (or any
provider) into the harness, this adds **provider bindings** — model
providers registered as data.

## What it does

A `Binding::Provider` record — `{ name, base_url, secret, wire format,
auth scheme, models, cost_usage_path }` — describes a gateway or vendor
endpoint entirely as configuration. Model bindings reference a provider;
resolution merges its base URL, credential, and declared wire format
(authoritative over the model-name/base-URL heuristics, which remain
unchanged as the fallback for the existing OpenAI/Anthropic/OpenRouter
paths — this PR is purely additive). Mirrored in the TS runtimes.

```
exo secret set opper --env OPPER_API_KEY
exo model provider register opper \
  --base-url https://api.opper.ai/v3/compat \
  --secret opper --format chat-completions --discover \
  --cost-usage-path opper.cost.total
exo model register terra --provider opper --model openai/gpt-5.6-terra
exo repl --model terra
```

Design points, following the review feedback rounds:

- **Dynamic registration**: `--discover` fetches the model list from the
  standard OpenAI-compatible `GET {base_url}/models` endpoint (using the
  provider's declared auth scheme); `--models a,b,c` declares it manually.
- **Auth is independent of wire format** (`--auth bearer|x-api-key|none`,
  defaulting to the format's native scheme) — e.g. Opper's
  Anthropic-compatible endpoint is Bearer-only. A provider either carries
  its own secret or is explicitly unauthenticated; ambient env keys are
  never used, and the TS runtimes suppress unwanted auth headers
  explicitly (null default headers) so exactly the declared credential is
  sent.
- **One base-URL contract**: records store the provider root; each client
  appends its own path (the Rust router joins `v1/messages`, the
  Anthropic SDKs append `/v1/...` themselves). `/v1`-suffixed anthropic
  roots are rejected in `Binding::validate()` at `put_binding`, so the
  CLI, HTTP API, and harness protocol all get the same error.
- **Cost as data**: cost is not part of the standard chat-completions
  usage schema — OpenRouter extends it at `usage.cost`, Opper at
  `usage.opper.cost.total` — so the record declares where to read it
  instead of hardcoding any vendor's extension. Applied in both runtimes,
  including the Anthropic streaming path (the SDK strips unknown fields
  from the accumulated final message, so it's captured from the raw
  `message_start`/`message_delta` events).

## Verification

- Unit: resolution merge + override precedence, declared-format/auth
  dispatch (both runtimes), discovery parsing, cost paths (incl.
  SSE-event-level regression), binding validation, serde wire-shape.
- Live against Opper: discovery (625 models in one call), registration
  UX (allow-list suggestions), and full agent turns through **both** the
  Rust and TS harnesses from a single provider record — chat-completions
  and anthropic-format/Bearer.
- `cargo test --workspace --all-targets`, `cargo clippy` (1.95, as CI
  pins) and `pnpm check` are green.

## Open questions

1. **Allow-list semantics**: the `models` list validates (with
   close-match suggestions) at `exo model register` time; at request
   time the provider stays authoritative, so a stale snapshot never
   breaks existing bindings. If you'd rather have hard runtime
   enforcement in `resolve_model_binding`, it's a small addition — your
   call.
2. **Auto-registration**: should discovered models become directly
   usable (e.g. agents naming `opper:openai/gpt-5.6-terra`) without a
   per-model `exo model register`? That changes model-name resolution
   semantics, so I left it out; happy to follow up if you want it.
3. **AGENTS.md `Value` exception**: `extract_provider_cost` walks the
   usage JSON with a runtime-configured path, which can't be a static
   serde struct. It extends the function's pre-existing shape — flagging
   it explicitly per the "no direct `Value` parsing" rule.


ankrgyl:
hi! thanks for the contribution. I think we will want to make routers configurable (ideally as just configuration, or with traits if needed), so that each router does not have to be maintained within exo itself. because of that, we probably won't accept this contribution as-is. If you'd like to work on extensible routing, though, we'd be thrilled to collaborate on that.

Jose-Sabater:
@ankrgyl - new proposed PR after your request to have providers as data, let me know :)

hnshah:
Tested the provider-binding branch at `b5838b0` against a local Ollama 0.32.4 instance. This appears to cover the Ollama transport case cleanly.

Configuration used:

```bash
exo model provider register ollama \
  --base-url http://127.0.0.1:11434/v1 \
  --format chat-completions \
  --auth none \
  --discover
```

Discovery returned all 33 local models. I then registered and exercised these three bindings through the Rust basic harness:

- `ornith:9b-q6`
- `qwen3.5:9b`
- `gpt-oss:20b`

For each model I verified both:

1. A normal completion (`Reply with exactly OK`) returned `OK`.
2. A full tool round trip invoked the `shell` tool, received `TOOL_OK`, then returned the final assistant response `FINAL_OK`.

Observed TTFTs for the normal completion were approximately 3.8s, 27.4s, and 17.1s respectively. In the tool-loop test all three produced the post-tool final response correctly.

Additional verification:

- `cargo build -p exo` passed on Rust 1.95.
- `cargo test -p exo provider` passed (5 tests).
- `cargo test -p executor provider` passed (6 tests; 1 live Opper test ignored as expected).

This also avoids the earlier failure mode where Ollama-bound models were sent through the Responses wire format. Happy to turn this into a focused Ollama regression test if that would be useful for this PR.

Alexsun1one:
Question on placement before I send something adjacent to this.

A chunk of open-weight models (qwen family, mimo, some GLM/DeepSeek deployments) never emit native `tool_calls` — they write the call as text inside message content, Hermes style:

```
<tool_call><function=NAME><parameter=KEY>VALUE</parameter></function></tool_call>
```

Exo drops those, so out of the box the model looks like it can't call tools at all.

I have a small patch that parses them into `function_call` outputs when native `tool_calls` are absent. Right now it sits in `responses.ts` as a fallback inside `chatCompletionToResponse`. But this PR turns providers into data with an explicit format field, and a text-encoded tool call is really a dialect of the same wire, so it might belong there instead.

Happy to send it either way — just want to aim it at the right place rather than rebase it later.

PR #177, opened 2026-07-27
Add ExoWorker autonomous harness example

Bautista1999:
## Summary
- add ExoWorker as a self-contained autonomous harness example with task-tree planning, durable memory, skills, adapters, scheduling, and sandbox tooling
- keep the example host-agnostic while allowing deployments to inject host tools through `registerHostTools`
- add dedicated setup/launch scripts, reasoning-safe history replay, documentation, and a live E2E smoke test

## Test plan
- [x] `pnpm check`
- [x] 20 test files / 177 tests pass
- [x] live local-process conversation turn and follow-up verified

Made with [Cursor](https://cursor.com)


examples/exo-worker/scripts/exoclaw-control
akrentsel, line 1: nit: mind renaming this to not commit in exoclaw naming? (and the other script here also named the same

examples/exo-worker/adapters/discord/discord.ts
```diff
@@ -0,0 +1,201 @@
+import type { AdapterAttachment, WorkerInboundEvent } from "../protocol";
```
akrentsel: these adapters are identical to Exo's, right? how about we show the re-use path here as a good example?

I think you don't need these – registerAdapterTools(tools) lets one start adapters and the adapters are the ones in Exo's codepath already

examples/exo-worker/docs/SELF-CONTROL.md
```diff
@@ -0,0 +1,165 @@
+# Exoclaw Self-Control
+
+Exoclaw is meant to be a long-running agent that can understand, maintain, and evolve the system it is operating inside.
```
akrentsel: Can remove this redundant docs (matches what's in the Exo folder)

akrentsel (commented):
main comment is around re-use – I'd love for this linked example to show the re-use path cleanly, which shouldn't require duplicating the adapter code as far as I can tell!

Bautista1999:
Addressed the reuse feedback in 6697a3c:

- Deleted the duplicated `examples/exo-worker/adapters/` tree. ExoWorker keeps `registerAdapterTools(tools)`; library workers continue to resolve from `examples/exo/adapters/` via `TypeScriptHarness::exo_from_root`.
- Removed the `exoclaw-*` scripts / scheduler-runner fork / unused `guardian-tools.ts` — those already live under `examples/exo/`.
- Dropped the redundant `docs/SELF-CONTROL.md` and pointed at the shared Exo copy.

The example also now shows the reuse path instead of a second adapter tree.

examples/exo-worker/adapters/discord/discord.ts
Bautista1999: Yeah you were right. registerAdapterTools() started the adapters from exo's `examples/exo/adapters/...`  . Was essentially dead code. Removed it.

(sorry about this one, im not using exo's adapters yet so didn't know this was just dead code)

examples/exo-worker/docs/SELF-CONTROL.md
Bautista1999: removed

examples/exo-worker/scripts/exoclaw-control
Bautista1999: Renamed as well

akrentsel:
one last suggestion – how about organizing this a bit more to have the tools be in the tools folder? you have the fal tool in there but others are up above the tool folder? After that, should be good to go!

Bautista1999:
> one last suggestion – how about organizing this a bit more to have the tools be in the tools folder? you have the fal tool in there but others are up above the tool folder? After that, should be good to go!

Yeah you're right, specially if we want this to be exemplary. Good point! Reorganazing a bit.

Bautista1999:
Ok @akrentsel done. I did further cleanup on some other parts:
1. Organized every tool under /tools
2. Removed/ cleaned up leftovers from WorkerClaw and some ExoClaw comments and some logic that falled back to legacy workerclaw paths
3. Wired fal tools since they were not registered in the harness, and set them in 'false' by default

akrentsel (approved):
ok, lgtm to go in – might need some moving around when we refactor the codebase soon, but otherwise fine!

PR #181, opened 2026-07-27
setup: reuse existing checkout when run from inside the repo

akrentsel:
Running `bash setup.sh` from inside an already-cloned exo repo cloned a fresh copy into a new `exo/` subfolder instead of using the code already there.

Two fixes in `choose_install_dir`'s checkout detection:
- `is_exo_checkout` required `.git` to be a directory, which fails in git worktrees (`.git` is a file there). Now checks `-e`.
- Detection only worked at the repo root. New `exo_checkout_root` uses `git rev-parse --show-toplevel`, so setup run from any subdirectory reuses the containing checkout's root.

The curl-piped install flow outside a checkout is unchanged.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -404,7 +404,17 @@ missing_has() {
 
 is_exo_checkout() {
   local dir="$1"
-  [[ -d "$dir/.git" && -f "$dir/exo.sh" ]]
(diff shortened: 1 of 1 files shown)


akrentsel:
small local usability change – so proceeding without review here

Issue #184, opened 2026-07-28
Scheduler: durable state, slot-anchored fires, and migrations (keeping it a host service)


ad1269:
# Scheduler: durable state, slot-anchored fires, and migrations (keeping it a host service)

Following up on the durability discussion in #architecture with something concrete, scoped to what the tools/adapters alignment keeps: the scheduler stays a supervised host service with tool commands. Nothing here touches the deferred scheduler-as-adapter question. The goal is narrower: a missed or lost fire should be impossible-or-explicit, and scheduler state should move with the agent.

**Everything below except F is implemented in the companion PR, #185** (five severable commits: records → grid+policy → `@at` → fire→wake → semantics tests). This issue carries the reasoning; the PR carries the diff.

## What I found reading the current implementation

1. **Fires drift.** `mark_completed` sets `next_run_at = now + interval` — anchored to *completion* (~`scheduler_types.rs:145`), so `@every 1h` with a 10-minute task actually fires every ~70 minutes, accumulating forever.
2. **Missed slots are undefined behavior.** If the host is off through N due times, they're silently skipped — by accident, not decision. Nothing records that they were missed.
3. **State writes aren't atomic.** Task records go down with bare `fs::write` (~`scheduler_store.rs:112, :139`). A crash mid-write corrupts the record — while the adapter store two files over already does temp+rename for exactly this reason.
4. **The due-task claim is read-then-write.** `claim_due_tasks` (~`scheduler_store.rs:81`) loads, mutates, puts — no conditional write, no heartbeat. Two runners both claim; the PID lockfile is the only real guard.
5. **The fire→wake handoff is lossy.** If the process dies between running a task and delivering the conversation wakeup, the wake is gone — and machine-sent events have no human retry loop behind them.

## Proposal

**A. Slot-anchored schedules.** Next fire derives from the schedule's grid (`anchor + n·interval`), never from completion time. Kills drift structurally.

**B. Explicit missed-fire policy — semantics in the syntax.** Each task declares what a downed host means: `missed: skip | once | all` (default `once`: fire a single catch-up run, then re-anchor to the grid — what a human assistant would do after a closed laptop). No implicit at-least-once ambiguity; an agent reading the task record knows exactly what recovery does.

[...]

PR #185, opened 2026-07-28
scheduler: durable state, slot-anchored fires, one-shot @at, and a durable fire→wake handoff

ad1269:
Design context and fuller reasoning: #184.

Follow-up from the #architecture durability discussion. The scheduler stays a supervised host service; the due-task claim path is untouched (deferred to #113, below). Five commits, severable, each green on the full suite.

## Problem

1. Fires drift: `next_run_at = completion + interval`, so `@every 1h` with a 10-minute command actually fires every ~70 minutes, forever.
2. Missed slots are undefined: slots that pass while the host is down are silently absorbed; nothing records they existed.
3. Task/run records are written with bare `fs::write` — a crash mid-write corrupts the record. The adapter store already does temp+rename.
4. A crash between running a task and waking the conversation loses the wake: the work happened, the agent never hears about it.

## Change

**Drift** (commit 2) — fires derive from the schedule's grid, not completion time:

```
today:  9:00 ─ run 10m ─ next 10:10 ─ 11:20 ─ 12:30    (completion-anchored)
now:    9:00 ─ run 10m ─ next 10:00 ─ 11:00 ─ 12:00    (grid-anchored)
```

**Missed slots** (commit 2) — each task declares `missed: skip | once | all` (default `once`: one catch-up run, then re-anchor). The policy engages only past one slot behind, so a slightly-late poll fires under every policy and `once` matches today's behavior in the common case. `all` caps at 100 and keeps the newest slots.

**Records** (commit 1) — temp+rename writes; `schema_version` with a step-per-version migration on load. Unversioned records read as v0; newer-than-known versions are a hard error. `.exo/scheduled-tasks` becomes a versioned unit that can move with the agent.

**One-shots** (commit 3, severable) — `@at <rfc3339>`; fires once, transitions to a terminal completed state. Severability checked by reverting it on the tip and re-running the suite.

**Fire→wake** (commit 4):

```
today:  run command ──► wake conversation
                    └── crash here: wake lost, nobody resends

now:    run command ──► write fire record ──► wake ──► mark delivered
        startup:        redeliver unmarked fire records
```

The fire record carries the rendered prompt, so redelivery never re-runs the command. Dedupe on `(task, slot)`; a delivered slot refuses re-queueing.

## Semantics

- A command that runs longer than its own interval serializes back to back — one slot behind at a time, nothing skipped, nothing overlapped (pinned in tests, commit 5). Overshooting by more than two intervals engages the missed policy.
- Follow-up worth discussing: surface overruns in the wakeup prompt so the agent can reschedule itself, rather than the harness picking a coping policy.
- `missed` is settable from Rust but not yet exposed in `scheduler-tools.ts` (closed property set; didn't want to ship unverified edits to the strict schema). One-file follow-up.

## Relationship to #113

- The claim race in `claim_due_tasks` is untouched: the right fix is a conditional write per occurrence, and that should follow the coordinator's direction rather than race it. @ankrgyl — happy to build it either shape once #113 settles; a comment at the call site says the same.
- Commit 4 and #113's queued report turns close the same lost-wake hole. When the coordinator lands this should collapse into the enqueued turn, and `(task, slot)` maps onto occurrence dedupe keys — happy to do that rebase. One distinction maybe worth keeping: an occurrence re-fire re-executes the command; the fire record redelivers existing output.
- Commits 1–3 are orthogonal to the coordinator; diffs kept small so rebasing over it stays trivial.

## Tests

Workspace 238 → 262. Grid math and policies are pure functions of record + clock — no sleeps. Each commit green on `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --all-targets`. Caveat: not yet exercised end-to-end against a live agent — automated coverage only so far; I'll follow up after a live pass.


61cygni:
This is a much needed change. In general looks good. However, we should definitely update the docs and model facing prompts. I'd recommend the following in this or a follow on PR:

- examples/exo/scheduler-tools.ts : Document @ one-shot schedules in the prompt, otherwise the model will have to guess at it looking at the source, Add strict-schema missed argument: "skip" | "once" | "all" | null, included in required.

- examples/exo/harness.ts : Tell the agent about @at, grid-anchored recurrence, missed-fire policies, and completed_at_ms.

- we should update website/docs-src/concepts/task-scheduler.md general documentation (having Fable or Sol update is fine I think), and docs/EXO-BASICS.md at a minimum

- it looks to me like the Rust backend accepts missed, but the closed TypeScript schema rejects it(?)

Otherwise this generally looks good.

ad1269:
Addressed all four, thanks for catching!

- scheduler-tools.ts:
    - schedule now documents `@at <rfc3339>` one-shots and grid anchoring alongside `@every`
    - `missed` added to the strict schema (`skip | once | all + null`, in required, matching sandboxMode's idiom — null maps to the once default).
    - The handler already forwarded args wholesale, so this also closes the Rust-accepts/TS-rejects mismatch you spotted.
    - `list_scheduled_tasks` now names `missed / last_missed_fire / completed_at_ms`, which the records already serialize.

- harness.ts:
    - the scheduler prompt section gained three lines: `@at` and its terminal state, grid-anchored fires, and the missed trio with once default.

- Docs: `task-scheduler.md` gained "When a task fires" + "Durability" sections; EXO-BASICS.md extended.

- Validation: 121 TS tests, typecheck/lint/format, and 111 executor Rust tests green (Rust untouched).

ad1269:
Apologies for the slight delay. This was my first time actually running exo end-to-end (plus a busy week), and I wanted to be thorough and verify the behavior live rather than just trust the suite. Documenting my testing and what I observed here:

- Booted the canonical setup and scheduled tasks from chat. The model picked `missed: skip` on its own from the new schema description, and it landed correctly in the task record. So the schema → tool → Rust chain works end to end.
- Watched grid anchoring live with a `sleep 30 && date` task: next fire = slot + interval, not completion + interval. On main this task drifts ~30s per cycle; here it held its grid seconds exactly.
- Left the runners stopped for ~22 hours with two tasks (one skip, one once) as an A/B. On restart: the once task fired exactly one catch-up (stamped at the restart minute), the skip task fired zero and resumed on-grid — both with the missed evaluation recorded. No burst from either.
- An `@at` one-shot fired once at its scheduled second, went to `completed_at_ms`, stays listed, never fired again.
- The per-minute run ledger shows exactly one run per task per minute across the whole session — no double-fires anywhere.

One thing I noticed that's not this PR (executions were provably single; this is downstream): around ExoChat relay reconnects, some report messages were delivered ×2–×4. The adapter's outbound resend after reconnect has no dedupe on the receiving side. I'll include that observation in the delivery/priorities write-up I mentioned in the channel, since it's the same layer.

61cygni:
Thanks for addressing these comments. Passes my standard gauntlet of tests. Looks good. Merging.

PR #188, opened 2026-07-29
exo: file edit tools chosen per model family

j4redux:
This adds edit tools, str_replace/create_file and apply_patch, registered per model family.

- `examples/exo/edit-tools.ts` (+test): Adding the three tools. Matching runs host-side as pure functions (the sandbox only moves bytes), and writes are temp-file+rename with nothing written unless the whole patch resolves; a failed edit can't half-write a source file. Tests fake the sandbox - no docker in my current env :(
- `examples/typescript/model-family.ts` (+test): resolves the model family from upstream model id. Router aliases resolve to `unknown`, which gets str_replace.
- `examples/typescript/turn-loop.ts`:  threads the resolved model id into the `registerTools` hook, so family resolution costs no extra round trip.
- `examples/exo/harness.ts`: registers the tools; tool descriptions and error messages carry usage guidance.
- `website/docs-src/tutorials/game-emulator-integration.md`: updates the caller of the exported `registerExoTools`, which gained a required `model` param.

`pnpm check` green: lint 0 warnings, typecheck clean, 190 tests (69 new).

### DIFF
--- a/examples/exo/edit-tools.test.ts
+++ b/examples/exo/edit-tools.test.ts
@@ -1,0 +1,820 @@
+import { describe, expect, it } from "vitest";
+
+import type { SandboxProcess } from "@exo/harness";
+
+import {
+  applyHunks,
+  applyStrReplace,
+  createEditToolInstances,
+  editToolsForFamily,
+  parseApplyPatch,
+  readSandboxFile,
+  validatePath,
+  writeSandboxFile,
+} from "./edit-tools";
+
+describe("editToolsForFamily", () => {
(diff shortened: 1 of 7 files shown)


akrentsel:
thanks for this! fyi we are about to land a change to how tools are added/registered in PR #171 courtesy of @61cygni, so will want to wait till then for anything with adding tools. we can then discuss what bucket to put these tools into – they best fit into exo-official tooling but not part of the minimal minimal set

i'd love to measure how much the model-aligned tools help task completion...

PR #190, opened 2026-07-29
Stop detached process bridge pollers

alexr17:
## Context

Dropping a process bridge's wait future detached its output poller, which continued issuing `recv` long polls indefinitely. The output and stdin background tasks should not outlive the returned process.

## Description

- Return both background task handles and own them with a guard captured by the wait future.
- Abort both tasks when the wait future is dropped, completes normally, or returns either task's error.
- Cover cancellation of an in-flight `recv` request and verify that polling does not restart.

## Testing

- `cargo test -p exoharness --features basic-backend process_bridge`
- `cargo fmt --check`

### DIFF
--- a/crates/exoharness/src/sandbox_provider/process_bridge.rs
+++ b/crates/exoharness/src/sandbox_provider/process_bridge.rs
@@ -9,6 +9,7 @@ use futures::future::BoxFuture;
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::sync::{mpsc, oneshot};
+use tokio::task::JoinHandle;
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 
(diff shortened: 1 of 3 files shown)


crates/exoharness/src/sandbox_provider/process_bridge.rs
```diff
@@ -362,34 +365,44 @@ pub fn process_parts(client: Arc<dyn Client>) -> crate::SandboxProcessParts {
     }
 }
 
+struct AbortOnDrop([JoinHandle<()>; 2]);
```
alexr17: so i could also enable rt to use: https://docs.rs/tokio-util/latest/src/tokio_util/task/abort_on_drop.rs.html

alexr17: okay actually this is cleaner doing it

61cygni:
Good find. Actually it looks like there are two leaks. The one you pointed out on the poller. But also it looks like the stdin forwarder lives on forever after a normal exit. Thanks for the PR!

PR #193, opened 2026-07-30
Adds optional EXO_AGENT_TOOLS_DIR +  helper to split assistant tool calls

Bautista1999:
## Summary
1. Add `resolveAgentToolsDirectory()` so hosts can override where `install_agent_tool` modules live via `EXO_AGENT_TOOLS_DIR`
  - Default remains `.exo/agent-tools` when unset
  - Document the override in the ExoWorker README
2. Also adds helper to split assistant tool calls for Response's API for OpenAi

## Why
1. Multi-agent hosts sharing one exo root need isolated agent-tool directories. This is an optional env override, not a behavior change for existing single-agent setups.
2. OpenAI, as it is, it's unsupported in exo-worker: OpenAI rejects receiveing multiple parallel tool calls in one assistant message


akrentsel (approved):
This makes sense to me. We will probably want such a mechanism for the main branch too (the specifying of a distinct agent directory) – though I'm not sure it makes sense to come from an env variable. perhaps eventually better set from via a binding. 

in any case, this is fine for this branch – in main tool use changes a bit and got reorganized.

akrentsel:
(and sorry for the delay on the review – still in Africa teaching, back mid-next week!)

PR #194, opened 2026-07-30
Codex and Claude Code as executors for the exo agent

j4redux:
This makes Claude Code and Codex usable as executors for the Exo agent on the exoharness substrate. The PR is split into four independently reviewable commits.

1. Expose registry tools inside each native runtime

Codex receives registry tools through its dynamic-tools channel, the same mechanism originally built for the gated exo_shell bridge. exo_shell remains gated, and shell execution stays native to Codex inside the external sandbox.

Claude Code receives the tools through an in-process MCP server.

In both cases:

* Registry JSON schemas pass through unchanged.
* Tool execution remains on the host, where the registry lives.
* An empty registry preserves the existing preset behavior.

This adds @modelcontextprotocol/sdk, which is already present in the lockfile through claude-agent-sdk, so it does not introduce a new download.

2. Add the composition harness

The new harness routes:

* Anthropic models to Claude Code
* OpenAI models to Codex

The executor is selected from the upstream model ID. Unknown model families fail explicitly rather than falling back silently.

Exo-provided tools, memory, scheduler, adapters, guardian, todos, skills, and web, are passed through the bridge. Shell execution remains native to the selected executor.

No Rust changes are required. The harness already runs with:

--harness examples/exo/coding-executor-harness.ts

using the executor sandbox image with networking enabled.

3. Make approval behavior explicit

Executor options now accept:

approvals: "auto" | "deny"

The default is "auto", treating the external sandbox as the containment boundary, consistent with the existing executor presets.

This also fixes two silent failure modes:

* Codex fileChange approval requests were falling through to the unhandled-request fallback and being declined.
* Claude Code had no permission handler, so permission prompts stalled the CLI until the startup timeout.

Codex approval decisions are emitted as codex_approval events.

Claude Code denials are emitted as claude_tool_permission events. Allowed tool calls are already visible through the existing observed-tool events.

4. Make Codex warm sessions rewind-safe

Each warm Codex session records the newest sandbox_started event ID when it is created.

Before reuse, the session checks that epoch. If a rewind has moved the sandbox to a different epoch, the session invalidates itself instead of passing the next turn to a dead process.

### Open questions

**Should fileChange be accepted under "auto"?**

Previously, fileChange requests were silently declined.

Because command execution is already accepted inside the external sandbox, declining file changes does not add meaningful protection. I treated the previous behavior as an accidental consequence of the fallback handler.

Please flag this if the decline was intentional.

**Should Claude Code use canUseTool or bypassPermissions?**

bypassPermissions would be simpler and would align with:

* Codex auto-accepting approvals
* The Claude SDK sandbox being disabled in favor of the external sandbox

I used canUseTool because the first commit changes the trust boundary: injected mcp__exo__* tools execute on the host and can perform actions such as guardian_action and send_adapter_message.

The callback therefore remains the only chokepoint between the model and host-side effects. It is also required for "deny" mode and provides the seam for a future interactive approval flow.

Switching "auto" to bypassPermissions would be a three-line change.

### Known limitations

* Codex warm threads retain the instructions and tools they were created with because the protocol does not support per-turn updates.
* A rewind triggered by another conversation on an agent-scoped sandbox is not visible to the current epoch check. That case still follows the existing error-then-recover path. Fixing it requires an agent-scoped event query in the TypeScript API.
* The approval policy currently lives in the harness options. Adding it to AgentConfig and exposing a CLI flag should follow once the policy shape is settled, rather than persisting a premature two-value enum in agent records.

**Validation**

pnpm check passes at every commit:

* Lint: 0 warnings
* Typecheck: clean
* Tests: 146 passing, including 25 new tests

Coverage includes type-level checks, unit tests, and a real MCP client/server exchange over an in-memory transport.

I have not yet run live Claude Code or Codex executions as I am currently afk!

### DIFF
--- a/examples/exo/coding-executor-harness.test.ts
+++ b/examples/exo/coding-executor-harness.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from "vitest";
+
+import { createToolRegistry, type TurnContext } from "@exo/harness";
+
+import {
+  executorForFamily,
+  registerCodingExecutorTools,
+} from "./coding-executor-harness";
+
+function fakeContext(enableAgentToolCreation = false): TurnContext {
+  return {
+    agentConfig: { typescript: undefined, enableAgentToolCreation },
+    conversationConfig: {},
+  } as unknown as TurnContext;
+}
+
+describe("executorForFamily", () => {
+  it("routes each family to its executor", () => {
+    expect(executorForFamily("anthropic")).toBe("claude-code");
+    expect(executorForFamily("openai")).toBe("codex");
+    expect(executorForFamily("unknown")).toBeNull();
+  });
+});
+
+describe("registerCodingExecutorTools", () => {
+  it("registers the exo agent tools but not shell", async () => {
+    const context = fakeContext();
+    const tools = createToolRegistry(context);
+    await registerCodingExecutorTools(tools, context);
+    expect(tools.get("shell")).toBeUndefined();
+    for (const name of [
+      "remember",
+      "todowrite",
+      "web_search",
+      "rewind_sandbox",
+      "rebuild_and_restart_exo",
+      "manage_tool",
+    ]) {
+      expect(tools.get(name), name).toBeDefined();
+    }
+  });
+
+  it("includes the agent-tool-creation built-ins when enabled", async () => {
+    const context = fakeContext(true);
+    const tools = createToolRegistry(context);
+    await registerCodingExecutorTools(tools, context);
+    expect(tools.get("install_agent_tool")).toBeDefined();
+    expect(tools.get("shell")).toBeUndefined();
+  });
+});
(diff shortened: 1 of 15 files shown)


Prithvi1994:
Is this direction still wanted?

Reading the executor/harness split, this looks like the piece that makes coding-agent executors usable *without* losing exo's tool surface. As things stand today, `--harness claude-code` registers no exo tools at all — no `manage_tool`, no `rebuild_and_restart_exo` — so the agent loses its self-maintenance path when it swaps executor.

Two things I noticed while reading:

- The branch is conflicted against `main` now (`mergeable_state: dirty`; base `372602d1`, `main` is `b06869a`).
- The two policy questions in the description — `fileChange` under `auto`, and `canUseTool` vs `bypassPermissions` — appear unanswered.

If this is still the plan, I'm happy to help: rebase onto `main`, propose answers to the two policy questions, or take a smaller slice if you'd rather split it. If it's been superseded by another approach, no problem — just point me at what replaced it.

Thanks for the detailed write-up; the approval-mode and warm-session sections in particular were useful.

PR #195, opened 2026-08-01
Generalize external sandbox providers

ankrgyl:
Remove the SandboxProvider enum so that users can create their own sandbox integrations without committing them in exo

### DIFF
--- a/crates/exoharness/src/types.rs
+++ b/crates/exoharness/src/types.rs
@@ -647,42 +647,45 @@ impl SandboxAttachment {
     }
 }
 
-#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize, PartialEq, Eq, Hash)]
-#[serde(rename_all = "snake_case")]
-pub enum SandboxProvider {
-    #[default]
-    Daytona,
-    E2b,
-    Sprites,
-    Vercel,
-    #[serde(rename = "aws_agentcore", alias = "aws-agentcore")]
-    AwsAgentCore,
-    AppleContainer,
-    Docker,
-    External,
-    #[serde(alias = "local")]
-    LocalProcess,
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(try_from = "String", into = "String")]
+pub struct SandboxProvider(Cow<'static, str>);
+
+impl Default for SandboxProvider {
+    fn default() -> Self {
+        Self::Daytona
+    }
 }
 
+#[allow(non_upper_case_globals)]
 impl SandboxProvider {
-    pub fn is_local(self) -> bool {
-        matches!(
-            self,
-            Self::AppleContainer | Self::Docker | Self::LocalProcess
-        )
+    pub const Daytona: Self = Self::from_static("daytona");
+    pub const E2b: Self = Self::from_static("e2b");
+    pub const Sprites: Self = Self::from_static("sprites");
+    pub const Vercel: Self = Self::from_static("vercel");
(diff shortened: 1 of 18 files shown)


crates/exoharness/src/types.rs
```diff
@@ -647,42 +647,45 @@ impl SandboxAttachment {
...
-    #[serde(alias = "local")]
-    LocalProcess,
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(try_from = "String", into = "String")]
+pub struct SandboxProvider(Cow<'static, str>);
+
+impl Default for SandboxProvider {
+    fn default() -> Self {
+        Self::Daytona
+    }
```
ankrgyl: this shouldn't be the default (i think we can remove the trait)

crates/exoharness/src/types.rs
```diff
@@ -647,42 +647,45 @@ impl SandboxAttachment {
...
-            Self::Sprites => "sprites",
-            Self::Vercel => "vercel",
-            Self::AwsAgentCore => "aws-agentcore",
-            Self::AppleContainer => "apple-container",
-            Self::Docker => "docker",
-            Self::External => "external",
-            Self::LocalProcess => "local-process",
+    pub fn is_local(&self) -> bool {
+        self == &Self::AppleContainer || self == &Self::Docker || self == &Self::LocalProcess
+    }
```
ankrgyl: this should be a property of the trait not of the name

crates/exoharness/src/types.rs
```diff
@@ -647,42 +647,45 @@ impl SandboxAttachment {
...
+
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    fn serialized_name(&self) -> &str {
+        match self.as_str() {
+            "aws-agentcore" => "aws_agentcore",
+            "apple-container" => "apple_container",
+            "local-process" => "local_process",
```
ankrgyl: this is dumb.. if we're worried about back-compat then we can just name these with underscores

crates/exoharness/src/types.rs
```diff
@@ -697,18 +700,35 @@ impl FromStr for SandboxProvider {
...
-            "vercel" => Ok(Self::Vercel),
-            "aws-agentcore" | "aws_agentcore" => Ok(Self::AwsAgentCore),
-            "apple-container" | "apple_container" => Ok(Self::AppleContainer),
-            "docker" => Ok(Self::Docker),
-            "external" => Ok(Self::External),
-            "local" | "local-process" | "local_process" => Ok(Self::LocalProcess),
-            provider => Err(anyhow::anyhow!("unsupported sandbox provider: {provider}")),
+        let provider = value.trim().replace('_', "-").to_ascii_lowercase();
+        let provider = match provider.as_str() {
+            "local" => "local-process".to_string(),
```
ankrgyl: no

crates/exoharness/src/types.rs
```diff
@@ -697,18 +700,35 @@ impl FromStr for SandboxProvider {
...
+            _ => provider,
+        };
+        if provider.is_empty()
+            || provider.starts_with('-')
+            || provider.ends_with('-')
+            || !provider
+                .bytes()
+                .all(|byte| byte.is_ascii_lowercase() || byte.is_ascii_digit() || byte == b'-')
+        {
+            return Err(anyhow::anyhow!("invalid sandbox provider: {value}"));
```
ankrgyl: no...

PR #196, opened 2026-08-02
discord: give every message create an enforced nonce

ad1269:
Redelivery after a worker restart re-sends anything un-acked, including sends that already landed. On Discord this is closable platform-correctly, so this PR does exactly that and nothing else.

## Change

Every message create carries Discord's idempotency key: `nonce` — a truncated hash of the command id and the chunk index — with `enforceNonce: true`. A redelivered command reuses its id, derives the same nonce, and Discord returns the existing message instead of posting a second one; it lands as an ordinary success and is acked like any other send. The nonce is per chunk because a long reply is several creates that must each survive — a shared nonce would swallow every chunk after the first. Attachments ride the first chunk, as they always have.

The worker's ack/nack behavior is untouched; the only change is how create payloads are constructed.

## Tests

138 → 146 on `pnpm check`. The nonce is deterministic across restarts, exactly 25 characters, distinct per chunk and per command, collision-free across the id/part boundary, and enforced on every payload; files land on chunk one.

Caveat: Discord documents the enforcement window only as "the past few minutes" — this is a replay cache, not a durable idempotency key. It cleanly covers the common case (a restarted worker replays its inflight within seconds); a retry delayed toward the runtime's five-minute backoff ceiling may land outside the window and duplicate. Deduplication is keyed on author + nonce and is server-side state, so it survives reconnects and new sessions; a deduplicated create returns the prior message as an ordinary success.

### DIFF
--- a/examples/exo/adapters/discord/discord.test.ts
+++ b/examples/exo/adapters/discord/discord.test.ts
@@ -4,11 +4,67 @@ import {
   attachmentKindForContentType,
   createResilienceHandlers,
   describeCloseCode,
+  discordMessagePayloads,
+  discordNonce,
   errorMessage,
   inboundAttachments,
   splitDiscordContent,
   startConnectionWatchdog,
 } from "./discord";
+
+describe("discordNonce", () => {
+  const commandId = "0198c4f1-7b2a-7c3d-8e4f-5a6b7c8d9e0f";
+
+  it("is deterministic across restarts", () => {
+    // A retry after a crash must derive the same nonce as the attempt that
+    // crashed, or Discord sees an unrelated message.
+    expect(discordNonce(commandId, 0)).toBe(discordNonce(commandId, 0));
+  });
+
+  it("fits inside Discord's 25 character limit", () => {
+    for (const part of [0, 1, 42, 1000]) {
+      expect(discordNonce(commandId, part).length).toBe(25);
+    }
+  });
+
+  it("differs per chunk so a long message is not swallowed", () => {
+    const nonces = [0, 1, 2].map((part) => discordNonce(commandId, part));
+    expect(new Set(nonces).size).toBe(3);
+  });
+
+  it("differs per command", () => {
+    expect(discordNonce("command-a", 0)).not.toBe(discordNonce("command-b", 0));
+  });
+
+  it("does not collide across the id/part boundary", () => {
+    expect(discordNonce("a-1", 0)).not.toBe(discordNonce("a", 10));
+  });
+});
+
+describe("discordMessagePayloads", () => {
+  it("puts an enforced nonce on every chunk", () => {
+    const payloads = discordMessagePayloads("cmd-1", ["one", "two"], []);
+    expect(payloads.map((payload) => payload.content)).toEqual(["one", "two"]);
+    for (const payload of payloads) {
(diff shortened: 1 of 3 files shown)


ad1269:
Cut down to the nonce, per the Discord thread. Writing down the state of the outbound path and the option space while it's fresh — mostly so it's on the record for whenever the durable-backend design happens.

**State of things.** Outbound is at-least-once: claim → dispatch → worker sends → ack; anything un-acked at a restart replays. There are two distinct crash windows in that path. A crash *mid-send* is genuinely ambiguous — no one can know whether the platform got it, so re-sending is the only correct move, and no sender-side mechanism changes that. A crash *after* the send completed but before the ack is consumed is different — the worker knew, and the knowledge died with it. Replays in the second window are guaranteed duplicates. Both are easy to hit deterministically (the earlier revision's IRC test: kill after send, restart, replay → duplicate, every time), and nothing sequences restarts against in-flight sends today — the irc-family workers exit on socket close by design, the adapter loop restarts on processing errors, a rebuild restarts everything, and shutdown has no drain step. What's *not* established is the real-world rate: the ×2–×4 I hit was on exochat during churn-heavy testing and I never root-caused it to a class. I'd expect rate to scale with traffic × restarts, but that's a projection.

**The option space:**
1. **Tolerate.** Fine for cosmetic chat double-posts; the de facto answer today. Stops being fine if adapters ever carry side-effectful sends.
2. **Drain on shutdown.** Stop claiming, let in-flight acks land, then kill. Cheap; covers planned restarts; can't cover crashes or disconnect-exits.
3. **Platform idempotency keys.** Best where they exist — the platform recognizes the retry itself, which covers even the ambiguous window (within the platform's enforcement window; Discord documents its own only as "the past few minutes"). Discord is the only one of the seven platforms here with a key; that's this PR.
4. **A sent-record at the worker.** No infrastructure; covers every restart class's post-send window; costs a per-worker contract. (The earlier revision of this PR; parked on my fork.)
5. **A durable delivery layer all workers submit through, with an idempotency key on submission.** Worker deaths still produce duplicate *submissions* — but they're deduped at the layer, so they stop mattering. The edge question doesn't disappear, it relocates: the layer's platform send and its record of that send are still two events, so the same window exists on its hop — but it fires at that component's restart rate instead of at hot-editable-worker restart rate, which likely rounds to never. Platform keys (3) still slot in at that layer for the platforms that have them.

**Which shape the backend takes decides how these compose.** If it's (5) — a delivery service the workers submit through — the worker-death class is handled by architecture and the residual is negligible-by-relocation. If it's a durable store *behind* the current architecture (workers still talking to platforms directly), the workers remain the effectors, the un-acked-send question stays at the worker, and it ends up combining with (2) or (4) anyway. Asked in Discord which of these you're picturing.

Meanwhile: this PR is now just the nonce, which covers the restart-replay case on Discord — a redelivered command reuses its id, derives the same nonce, and Discord returns the existing message — with the caveat that Discord's window is a few-minutes replay cache, so a retry delayed toward the backoff ceiling may fall outside it. And I'm running the sent-record branch on my own setup, where its dedupe events double as a counter: if normal usage shows a rate worth caring about, I'll bring back a number.

61cygni:
This is great. Thanks for doing.

PR #198, opened 2026-08-05
Keep ExoWorker prompts inside model context windows

Bautista1999:
## Summary
- Cap tool-result payloads when materializing history (32K for the latest round, 4K for older results).
- Add model-aware context budgeting: resolve known windows, learn real limits from provider errors, and compress middle history while pinning instructions, the client request, recent turns, and tool_call/tool_result pairs.
- Retry after context-window failures with a forced compress pass.
- Bound trailing vision attaches (max images and decoded bytes per round) so screenshots cannot inflate prompt estimates or blow the window.
- Emit host-neutral compression markers and prompt-usage custom events for optional UI mirroring.

## Test plan
- [x] Run ExoWorker unit tests for message-materialize and context-compress
- [x] Confirm a long tool-heavy turn compresses before the model call instead of failing with a context-window 400
- [x] Confirm a context-window 400 triggers force-compress and a successful retry
- [x] Confirm oversized or excess trailing screenshots are omitted with a text note rather than attached
- [x] Confirm tool_call / tool_result pairs stay intact across compression


examples/exo-worker/context-window.ts
```diff
@@ -0,0 +1,167 @@
...
+export const DEFAULT_CONTEXT_WINDOW_TOKENS = 200_000;
+
+/**
+ * Known prompt/context windows (tokens). Keys are bare API model ids or
+ * common prefixes; matching is substring / exact after stripping provider.
+ *
+ * Keep in sync with the host app's worker-model catalog when one exists
+ * (ExoWorker ships standalone and does not import host packages).
+ */
+export const KNOWN_CONTEXT_WINDOWS: Readonly<Record<string, number>> = {
```
akrentsel, line 19: ideally these would be pulled from somewhere that is kept up to date. see how we pull costs for cost accounting.

akrentsel (approved):
Looks fine to me!

PR #202, opened 2026-08-07
Add trial adapter and Harbor evaluation integration

akrentsel:
Adds the generic trial adapter and the Harbor evaluation workflow built on it. Includes trial lifecycle acknowledgements, partial-trajectory export, cancellation, snapshots, feedback reflection, task filtering, and smoke datasets.

The commits remain separated by layer for review.

### DIFF
--- a/crates/executor/src/adapter/runtime.rs
+++ b/crates/executor/src/adapter/runtime.rs
@@ -15,6 +15,7 @@ use lingua::universal::{TextContentPart, UserContent, UserContentPart};
 
 use super::store::{AdapterStore, stable_target_key};
 use super::tools::download_attachment;
+use super::trial::prepare_trial_run;
 use super::types::{
     AdapterAttachment, AdapterAttachmentKind, AdapterConfig, AdapterDeliveryStatus,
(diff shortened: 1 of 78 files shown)


crates/executor/src/adapter/runtime.rs
```diff
@@ -585,6 +586,9 @@ async fn handle_worker_event(
                 &metadata,
             )
             .await?;
+            if config.adapter_type == "trial" {
```
akrentsel, line 622: This is a bit ugly to have in here, in my opinion. Thinking through how to elegantly get around this issue... need to be able to set the sandbox here (via attach) before the convo starts.

exo/scripts/deferred-rebuild-and-restart
```diff
@@ -12,6 +12,7 @@
 #   $4 reason_json          JSON-encoded string|null
 #   $5 exo_bin
 #   $6 env_file
+#   $7 exo_root
```
akrentsel, line 15: note: all changes in this file stem from the fact that

(1) these eval runs have their own dedicated root directories
(2) you may run other exos on the same machine at the same time (I have my canonical running Exo going, at the same time as the eval one) so don't want their restarts to interfere.

PR #204, opened 2026-08-09
Add Harbor trial evaluation integration

akrentsel:
Adds the Harbor bridge and evaluation runner over the generic trial adapter, including ATIF trajectory export and local smoke datasets.

Stacked on #202.


akrentsel:
Note: tweaking the prompts still, but have been focused on wiring together the infra correctly. This is finally the correct shape for the evals. As they run overnight, will see whether the agent is properly using its capabilities during runtime.

akrentsel:
Superseded by the consolidated implementation in #202.

PR #206, opened 2026-08-11
exochat: replay-protect sends with a bounded relay cache

ad1269:
This gives exochat what #196 gave Discord: a platform-checked idempotency key. Discord already had one (`enforceNonce`) and we just started sending it; exochat is our own platform, so this adds the key on the relay side and uses it from the adapter.

**What it does.** The adapter stamps each outgoing envelope with an opt-in `id` (sha256 of the command id — the relay never sees raw command ids). The relay broadcasts once, records the id, acks. A replayed id — worker crashed between send and ack, restarted, retried — gets a duplicate ack and no second broadcast. And `command_ack` now waits for the relay ack, so an unacked send fails and retries instead of being assumed delivered.

**State and cost.** Per channel: at most 64 ids for at most 10 minutes, written through to durable-object storage (the WS hibernation API drops in-memory state between messages, so a plain ring buffer forgets). Typical size ~2–4 KB; adversarial hard cap ~10 KB if a client sends max-length ids. Cleanup is the existing 30-minute session alarm — no new lifecycle, no new cleanup path.

**Wire.** The protocol had no ack of any kind; the new `exo.chat.ack` frame is plaintext relay metadata, like presence. The `id` rides outside the AAD, which is what keeps old peers wire-compatible — but it also means the id is unauthenticated until an envelope version bump (a forged or stripped id can cause a dropped or duplicate delivery, nothing worse; the ciphertext path is untouched). Deploy relay before clients. Old clients are unaffected: id-less traffic takes the legacy path byte-for-byte and is never acked.

**What it is not.** Not exactly-once. If the window is gone — TTL expired, session torn down, storage lost — a replay is delivered again as an honest duplicate. At-least-once holds in every failure I could construct; the cache narrows the duplicate window, it can't close it.

**Tests.** 17 new — crash-between-send-and-ack replay, TTL expiry, cache loss, pre-id peer compatibility — with the real adapter worker as a subprocess against the real relay class. Core diff is ~186 lines; the rest is testkit and tests. End-to-end evidence, including a run on real workerd via `wrangler dev`, is in the PR thread.

### DIFF
--- a/exo/adapters/exochat/exochat.test.ts
+++ b/exo/adapters/exochat/exochat.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from "vitest";
+
+import { parseSendAck, sendDedupeId } from "./exochat";
+
+describe("sendDedupeId", () => {
+  const commandId = "0198c4f1-7b2a-7c3d-8e4f-5a6b7c8d9e0f";
+
+  it("is deterministic across restarts", () => {
+    // A replayed command must present the same id as the attempt that
+    // crashed, or the relay sees an unrelated send and broadcasts again.
+    expect(sendDedupeId(commandId)).toBe(sendDedupeId(commandId));
+  });
+
+  it("differs per command", () => {
+    expect(sendDedupeId("command-a")).not.toBe(sendDedupeId("command-b"));
+  });
+
+  it("does not leak the command id to the relay", () => {
+    expect(sendDedupeId(commandId)).not.toContain(commandId);
+  });
+
+  it("fits the relay's id bounds", () => {
+    const id = sendDedupeId(commandId);
+    expect(id.length).toBeGreaterThanOrEqual(8);
+    expect(id.length).toBeLessThanOrEqual(128);
+  });
+});
+
+describe("parseSendAck", () => {
+  it("accepts a relay ack", () => {
+    expect(
+      parseSendAck({
+        channel: "exo.chat.ack",
+        id: "dedupe-id-1",
+        duplicate: true,
+        at: 123,
+      }),
+    ).toEqual({ id: "dedupe-id-1", duplicate: true });
+  });
+
+  it("rejects envelopes, presence, and malformed acks", () => {
+    expect(parseSendAck({ channel: "exo.chat", id: "dedupe-id-1" })).toBeNull();
+    expect(
+      parseSendAck({ channel: "rendezvous", type: "presence" }),
+    ).toBeNull();
+    expect(parseSendAck({ channel: "exo.chat.ack", id: 7 })).toBeNull();
+    expect(parseSendAck(null)).toBeNull();
+    expect(parseSendAck("exo.chat.ack")).toBeNull();
+  });
+});
--- a/exo/adapters/exochat/exochat.ts
+++ b/exo/adapters/exochat/exochat.ts
@@ -1,0 +1,33 @@
+import { createHash } from "node:crypto";
+
+// The relay's replay-protection id rides at envelope level, outside the
+// ciphertext, so the relay can see it. Send a hash of the command id rather
+// than the id itself: deterministic across retries (a replayed command
+// presents the same key), meaningless to the relay.
+export function sendDedupeId(commandId: string): string {
+  return createHash("sha256")
+    .update(commandId)
+    .digest("base64url")
+    .slice(0, 22);
+}
+
+export type SendAck = {
+  id: string;
+  duplicate: boolean;
+};
+
+// Acks are relay metadata like presence, not encrypted frames: the relay has
+// no channel key, and the ack carries only the id the sender already exposed.
+export function parseSendAck(value: unknown): SendAck | null {
+  if (!value || typeof value !== "object" || Array.isArray(value)) {
+    return null;
+  }
+  const record = value as Record<string, unknown>;
+  if (record.channel !== "exo.chat.ack") {
+    return null;
+  }
+  if (typeof record.id !== "string" || typeof record.duplicate !== "boolean") {
+    return null;
+  }
+  return { id: record.id, duplicate: record.duplicate };
+}
--- a/exo/adapters/exochat/relay-testkit.ts
+++ b/exo/adapters/exochat/relay-testkit.ts
@@ -1,0 +1,352 @@
+// Test-only host for the real ExoChat relay: RendezvousSession from
+// website/src/worker.js served over a minimal RFC 6455 server, so the real
+// adapter process and a real peer client exercise the deployed relay code
+// end-to-end. The durable-object runtime is shimmed (sockets survive an
+// instance swap, storage is a map), which is exactly the hibernation contract
+// the replay cache is written against.
+import { createHash } from "node:crypto";
+import http from "node:http";
+import type { Duplex } from "node:stream";
+
+import type {
+  RelayContext,
+  RelayEnv,
+  RelaySocket,
+} from "../../../website/src/worker.js";
+import { RendezvousSession } from "../../../website/src/worker.js";
+
+const WS_GUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
+const WS_PATH_PATTERN = /^\/chat\/ws\/([A-Za-z0-9_-]{8,128})\/?$/;
+
+class MemoryStorage {
+  values = new Map<string, unknown>();
+
+  async get(key: string): Promise<unknown> {
+    return this.values.get(key);
+  }
+
+  async put(key: string, value: unknown): Promise<void> {
+    this.values.set(key, value);
+  }
+
+  async setAlarm(): Promise<void> {}
+
+  async deleteAll(): Promise<void> {
+    this.values.clear();
+  }
+}
+
+type HostedSocket = RelaySocket & {
+  attachment: { role?: string } | undefined;
+};
+
+type Channel = {
+  context: RelayContext;
+  session: RendezvousSession;
+  sockets: Set<HostedSocket>;
+  storage: MemoryStorage;
+  ackDropsLeft: number;
+  droppedAcks: string[];
+  ackDropWaiters: (() => void)[];
+};
+
+export type RelayHarness = {
+  url: string;
+  dropAcks(channelId: string, count: number): void;
+  waitForDroppedAck(channelId: string): Promise<void>;
+  restartChannel(
+    channelId: string,
+    options: { clearStorage: boolean },
+  ): Promise<void>;
+  close(): Promise<void>;
+};
+
+export async function startRelayHarness(env?: RelayEnv): Promise<RelayHarness> {
+  const channels = new Map<string, Channel>();
+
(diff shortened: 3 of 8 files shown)


ad1269:
## End-to-end evidence: exactly-once on the real relay runtime

Two layers of proof, run on this branch as it sits.

### 1 · The automated battery (testkit-hosted relay class)

`pnpm vitest run exo/adapters/exochat/replay.e2e.test.ts` — the real `worker.ts` adapter as a SIGKILLed-and-restarted child process, the real `RendezvousSession` class, a peer speaking the pre-id wire protocol:

```
✓ replays after a crash between send and ack without a second broadcast   643ms
✓ re-broadcasts honestly after the replay window expires                  648ms
✓ dedupes across a relay wake, and degrades to at-least-once when the
  cache is lost                                                           471ms
✓ relays id-less traffic from an old client untouched and never acks it   239ms

Test Files  1 passed (1) · Tests  4 passed (4)
```

The first test is the crash window this design exists for: the harness drops the relay's ack, the worker dies unacked, the restarted worker replays the same command id, and the peer sees exactly one broadcast. (Full suite: 163/163.)

### 2 · The same journey on the real runtime — wrangler dev, workerd, SQLite-backed Durable Object

The testkit shims the hibernation contract; this run removes the shim. Relay started exactly as the adapter README documents:

```
pnpm dlx wrangler dev --config website/wrangler.jsonc --port 8787
```

then the real `worker.ts` (the same process `exo.sh` runs) driven as a child process, with a pre-id-protocol peer counting decrypted broadcasts. Two kill interleavings:

**Kill at first broadcast** (ack raced ahead this run — recorded, not assumed):

```
[journey] worker#1 sends command 0198c4f1… : "exactly once, please"
[journey] peer ← broadcast #1: "exactly once, please" (envelope id MTqTbUCFKi…)
[journey] worker#1 → command_ack (command 0198c4f1…)
[journey] SIGKILL worker#1 at first broadcast — ack HAD reached it
[journey] worker#2 (same state dir) REPLAYS the same command id
[journey] worker#2 → command_ack (command 0198c4f1…)
[journey] RESULT: peer saw 1 broadcast(s); replay acked: yes
[journey] control: a NEW command id must still broadcast
[journey] peer ← broadcast #2: "a genuinely new message" (envelope id AI-ukkY31b…)
[journey] RESULT: peer saw 2 broadcasts total — dedupe is per command id, not a mute
[journey] JOURNEY PASSED: exactly-once held on the real relay runtime
```

**Kill in the same tick as the write** — the point of this variant is that the outcome is deterministic even though the interleaving isn't: either the frame reached the relay (replay dedupes) or it didn't (replay delivers), and both branches end at exactly one broadcast:

```
[journey] worker#1 sends command 0198c4f1… : "exactly once, please" (kill mode: write)
[journey] SIGKILL worker#1 in the same tick as the write — the frame may or
          may not have reached the relay
[journey] peer ← broadcast #1: "exactly once, please"
[journey] after the crash the peer has seen 1 broadcast(s) — either branch
          must end at exactly one
[journey] worker#2 (same state dir) REPLAYS the same command id
[journey] RESULT: peer saw 1 broadcast(s); replay acked: yes
[journey] JOURNEY PASSED: exactly-once held on the real relay runtime
```

(Transcripts trimmed of connection-lifecycle noise. In both same-tick runs the frame had in fact already flushed to the socket; the frame-never-arrived branch is the trivial one — nothing to dedupe, the replay is just the delivery — and the ack-dropped-by-the-relay window is the one the vitest suite pins deterministically via its drop-ack hook.)

Worth noting from the second transcript: the restarted worker's retry loop re-sent the old command id once more *after* its ack — the relay duplicate-acked it and the peer count stayed at 1, which is the window doing its job under real retry behavior, not just under scripted replay.

<details>
<summary>The journey driver (self-contained, ~150 lines — run from the repo root with the relay up)</summary>

```js
// Replay-protection journey against the REAL relay (wrangler dev / workerd,
// SQLite-backed Durable Object) using the REAL adapter worker (worker.ts).
// Usage:  node replay-journey.mjs          (kill at first broadcast)
//         KILL_AT=write node replay-journey.mjs   (kill in the write's tick)
import { spawn } from "node:child_process";
import crypto from "node:crypto";
import fs from "node:fs/promises";
import os from "node:os";
import path from "node:path";

const REPO = process.cwd();
const WORKER = path.join(REPO, "exo/adapters/exochat/worker.ts");
const TSX = path.join(REPO, "node_modules/tsx/dist/cli.mjs");
const BASE = "http://127.0.0.1:8787";
const CHANNEL = `journey-${crypto.randomBytes(4).toString("hex")}`;
const SECRET = crypto.randomBytes(32).toString("base64url");
const COMMAND_ID = "0198c4f1-7b2a-7c3d-8e4f-5a6b7c8d9e0f";

const log = (m) => console.log(`[journey] ${m}`);
const sleep = (ms) => new Promise((r) => setTimeout(r, ms));

async function waitUntil(get, what, timeout = 15000) {
  const deadline = Date.now() + timeout;
  for (;;) {
    const v = get();
    if (v !== null && v !== undefined && v !== false) return v;
    if (Date.now() > deadline) throw new Error(`timed out waiting for ${what}`);
    await sleep(25);
  }
}

class Adapter {
  constructor(stateDir, label) {
    this.events = [];
    this.label = label;
    this.child = spawn(process.execPath, [TSX, WORKER], {
      env: {
        ...process.env,
        EXO_ADAPTER_CONFIG: JSON.stringify({ baseUrl: BASE, channelId: CHANNEL, secret: SECRET }),
        EXO_ADAPTER_STATE_DIR: stateDir,
      },
      stdio: ["pipe", "pipe", "pipe"],
    });
    let rest = "";
    this.child.stdout.setEncoding("utf8");
    this.child.stdout.on("data", (chunk) => {
      rest += chunk;
      const lines = rest.split("\n");
      rest = lines.pop() ?? "";
      for (const line of lines) {
        if (!line.trim()) continue;
        const ev = JSON.parse(line);
        this.events.push(ev);
        log(`${this.label} → ${ev.type}${ev.command_id ? ` (command ${ev.command_id.slice(0, 8)}…)` : ""}`);
      }
    });
    this.child.stderr.resume();
  }
  async ready() {
    await waitUntil(() => this.events.some((e) => e.type === "connected"), `${this.label} connected`);
  }
  send(id, text) {
    this.child.stdin.write(`${JSON.stringify({ type: "send_message", id, text })}\n`);
  }
  acks() {
    return this.events.filter((e) => e.type === "command_ack").length;
  }
  kill() {
    this.child.kill("SIGKILL");
  }
}

function peerConnect() {
  const key = Buffer.from(
    crypto.hkdfSync("sha256", Buffer.from(SECRET, "base64url"), Buffer.from(CHANNEL),
      Buffer.from("exo-chat-relay:aes-gcm:v1"), 32),
  );
  const url = new URL(`/chat/ws/${CHANNEL}`, BASE);
  url.protocol = "ws:";
  url.searchParams.set("role", "user");
  const ws = new WebSocket(url);
  const state = { chats: [], ws };
  ws.addEventListener("message", (event) => {
    if (typeof event.data !== "string") return;
    const msg = JSON.parse(event.data);
    if (msg.channel !== "exo.chat" || msg.from !== "agent") return;
    try {
      const bytes = Buffer.from(String(msg.ciphertext), "base64url");
      const d = crypto.createDecipheriv("aes-256-gcm", key, Buffer.from(String(msg.nonce), "base64url"));
      d.setAAD(Buffer.from(JSON.stringify({
        channel: msg.channel, channelId: msg.channelId, from: msg.from,
        nonce: msg.nonce, seq: msg.seq, version: msg.version,
      })));
      d.setAuthTag(bytes.subarray(bytes.length - 16));
      const frame = JSON.parse(
        Buffer.concat([d.update(bytes.subarray(0, bytes.length - 16)), d.final()]).toString("utf8"),
      );
      if (frame.type === "chat") {
        state.chats.push(frame.text);
        log(`peer ← broadcast #${state.chats.length}: "${frame.text}" (envelope id ${String(msg.id).slice(0, 10)}…)`);
      }
    } catch {
      log("peer ← envelope FAILED authentication (would be a wire break)");
    }
  });
  return new Promise((resolve, reject) => {
    ws.addEventListener("open", () => resolve(state), { once: true });
    ws.addEventListener("error", () => reject(new Error("peer connect failed")), { once: true });
  });
}

const stateDir = await fs.mkdtemp(path.join(os.tmpdir(), "exochat-journey-"));
log(`relay: ${BASE} (wrangler dev, real workerd + SQLite DO) · channel: ${CHANNEL}`);

const peer = await peerConnect();
log("peer connected (speaks the PRE-id wire protocol — fixed-field AAD, no ack knowledge)");

let worker = new Adapter(stateDir, "worker#1");
await worker.ready();

const KILL_AT = process.env.KILL_AT ?? "broadcast";
log(`worker#1 sends command ${COMMAND_ID.slice(0, 8)}… : "exactly once, please" (kill mode: ${KILL_AT})`);
worker.send(COMMAND_ID, "exactly once, please");
if (KILL_AT === "write") {
  worker.kill();
  log("SIGKILL worker#1 in the same tick as the write — the frame may or may not have reached the relay");
  await sleep(750);
  log(`after the crash the peer has seen ${peer.chats.length} broadcast(s) — either branch must end at exactly one`);
} else {
  await waitUntil(() => peer.chats.length >= 1, "first broadcast");
  const ackedBeforeKill = worker.acks();
  worker.kill();
  log(`SIGKILL worker#1 at first broadcast — ack ${ackedBeforeKill ? "HAD" : "had NOT"} reached it (recorded, not assumed)`);
}

worker = new Adapter(stateDir, "worker#2");
await worker.ready();
log(`worker#2 (same state dir) REPLAYS the same command id`);
worker.send(COMMAND_ID, "exactly once, please");
await waitUntil(() => worker.acks() >= 1, "replay ack");
await sleep(750);

log(`RESULT: peer saw ${peer.chats.length} broadcast(s); replay acked: yes`);
if (peer.chats.length !== 1) {
  log("FAIL: duplicate reached the peer");
  process.exit(1);
}

log("control: a NEW command id must still broadcast");
worker.send("0198c4f1-ffff-7c3d-8e4f-5a6b7c8d9e0f", "a genuinely new message");
await waitUntil(() => peer.chats.length >= 2, "control broadcast");
await sleep(250);
log(`RESULT: peer saw ${peer.chats.length} broadcasts total — dedupe is per command id, not a mute`);

log(peer.chats.length === 2 ? "JOURNEY PASSED: exactly-once held on the real relay runtime" : "FAIL");
worker.kill();
peer.ws.close();
process.exit(peer.chats.length === 2 ? 0 : 1);
```

</details>

Remaining gap between this and production Cloudflare: `wrangler dev` runs real workerd with real DO storage but not the production hibernation eviction timing — the storage-survives-instance-death property that covers is exercised by the third vitest scenario (`restartChannel` with storage intact). The live end-to-end through `EXO_CHAT_BASE_URL=http://127.0.0.1:8787 ./exo.sh fresh` with a browser client is the one journey not in this comment.

PR #208, opened 2026-08-13
Run exo sandboxes as smolvm microVM

BinSquare:
adding smolvm as a "sandbox" backend

Here's a quick way to check:

```
cargo build -p exo

./target/debug/exo agent create myagent --model m --sandbox-provider smolvm \
  --sandbox-image alpine:latest --networking enabled

./target/debug/exo conversation create myagent c --sandbox-provider smolvm \
  --sandbox-image alpine:latest --shell-program /bin/sh

./target/debug/exo conversation sandbox run myagent c "uname -sr"
```

Screenshot 2026-08-13 at 5 57 35 AM


crates/exoharness/src/sandbox_provider/smolvm.rs
```diff
@@ -0,0 +1,1069 @@
...
+
+impl SmolvmSandboxBackend {
+    pub fn new() -> Self {
+        Self::with_mode(SmolvmExecutionMode::default())
+    }
+
+    pub fn with_mode(mode: SmolvmExecutionMode) -> Self {
+        let binary = std::env::var_os(SMOLVM_BIN_ENV)
+            .map(PathBuf::from)
+            .unwrap_or_else(|| PathBuf::from(DEFAULT_SMOLVM_BIN));
```
ankrgyl: we prefer to propagate variables like this through clap so that we don't end up with hidden tuning params (i.e. everything can be provided via struct)

also curious how did you think of the trade off between using this binary vs. embedding smolvm / calling the SDK directly?

crates/exoharness/src/sandbox_provider/smolvm.rs
```diff
@@ -0,0 +1,1069 @@
...
+/// parent-death watchdog. Ephemeral runs only: a persistent machine is meant to
+/// outlive the `machine start` that created it.
+///
+/// Prefers the sibling `smolvm-bin`, since a packaged `smolvm` may be a wrapper
+/// script that cannot be exec'd as the boot binary.
+fn resolve_boot_binary(binary: &Path) -> Option<PathBuf> {
+    if let Some(explicit) = std::env::var_os("SMOLVM_BOOT_BINARY") {
+        return Some(PathBuf::from(explicit));
+    }
+    let resolved = which_binary(binary)?;
```
ankrgyl: it would be nice to do this async if possible. or if not, then cache the value. otherwise we have a bash command that blocks the async worker that runs this for every operation

crates/exoharness/src/sandbox_provider/smolvm.rs
```diff
@@ -0,0 +1,1069 @@
...
+    }
+
+    /// A delete for a machine that is not there — the desired end state.
+    pub fn no_such_machine(stderr: &str) -> bool {
+        stderr.contains("not found") || stderr.contains("does not exist")
+    }
+}
+
+/// `(major, minor, patch)` from `smolvm --version`; `None` means "assume old".
+fn parse_version(output: &str) -> Option<(u32, u32, u32)> {
```
ankrgyl: can we use a semver library?

crates/exoharness/src/sandbox_provider/smolvm.rs
BinSquare: my thought process: bringing in embedded sdk means everyone using exo takes on a ~30mb dependency even if they wouldn't be using smol machines but the setup is minimal

vs calling the binary, user needs to do the setup if they want to use it from exo

ankrgyl: yeah true, but if it's technically superior to embed, we can always support via feature flag

BinSquare: It's the same tech under the hood~ so no difference in behavior at all.

Happy to support via flag if tighter integration is reasonable tho!

crates/exoharness/src/sandbox_provider/smolvm.rs
BinSquare: yes, will implement

BinSquare: implemented

crates/exoharness/src/sandbox_provider/smolvm.rs
BinSquare: implementing this as suggested with the async route

BinSquare:
@ankrgyl Mind taking another look?

crates/exoharness/src/sandbox_provider/smolvm.rs
```diff
@@ -0,0 +1,1187 @@
...
+    /// Probed once: re-asking per `acquire` would spawn a process per sandbox.
+    capabilities: OnceCell<Capabilities>,
+    /// Last use of each warm machine this process created, for TTL reaping.
+    warm_seen: Mutex<HashMap<String, Instant>>,
+}
+
+impl SmolvmSandboxBackend {
+    pub fn new() -> Self {
+        Self::with_mode(SmolvmExecutionMode::default())
+    }
```
ankrgyl, line 112: small nit: why not make this the implementation of `default()`?

ankrgyl (approved):

ankrgyl:
lgtm, i can land it once you reject or do the little nit i suggeseted :)

crates/exoharness/src/sandbox_provider/smolvm.rs
BinSquare: That's a good simplification, implementing

BinSquare: Landed

BinSquare:
> lgtm, i can land it once you reject or do the little nit i suggeseted :)

landed the refactor, PTAL!

PR #210, opened 2026-08-14
fix: make ./exo.sh fresh the obvious local wipe

### LINKED ISSUE #93: Simplify scripts/exo.sh reset/delete UX
## Problem

The current destructive reset flow is confusing. For example, `./scripts/exo.sh delall all` looks like two commands or a typo-prone incantation.

## Desired behavior

Expose one obvious command for clearing local Exo state, such as `reset`, `clean`, or `delete-all`.

The command should make the blast radius explicit before deleting anything.

## Acceptance criteria

- There is a single documented command for deleting all local dev state.
- The command explains exactly what will be deleted.
- Destructive behavior requires an explicit confirmation unless a clearly named force flag is used.
- README/setup docs stop recommending `delall all`.

gvkhosla:
## Summary

Fixes **#93**: `./exo.sh delall all` was the documented wipe and looked like a typo.

`./exo.sh fresh` is now the one command. It prints what will be deleted, then requires typing `fresh` unless `--force` is set. `delall all` / `delete-all all` / `reset` still work as aliases.

The confirmation is honest about blast radius: deleting an agent also deletes that agent's model bindings and stored secrets. `.env` and `.exo/exo-profile.md` are kept.

## Test plan

- [x] `bash -n exo.sh setup.sh`
- [x] `./exo.sh --help` documents `fresh` and `--force`, not `delall all`
- [x] `./exo.sh fresh </dev/null` refuses without `--force`
- [x] `./exo.sh delall` / `./exo.sh fresh extra` reject bad usage

Closes #93.


gvkhosla:
Rebased onto current `main` (no conflicts). `./exo.sh fresh` is still the wipe path; this just makes the confirmation honest about bindings/secrets and adds `--force`.

PR #211, opened 2026-08-14
fix: make missing-model errors actionable

### LINKED ISSUE #95: Make missing-model errors actionable
## Problem

When the default model is not registered, the user currently sees a low-level TypeScript harness failure and stack trace:

```text
Error: typescript harness failed: model is not registered: gpt-5.4; run `exo model register gpt-5.4 --secret <secret>`
```

This is technically accurate, but too noisy and incomplete for a normal user.

## Desired behavior

Missing-model failures should produce a short, actionable recovery message instead of a raw stack trace.

## Acceptance criteria

- The REPL catches missing-model errors and displays a concise explanation.
- The message includes the exact next commands or points to the bootstrap wizard.
- The stack trace is hidden by default for this expected setup failure.
- The behavior is covered for the default `./scripts/exo.sh` path.

gvkhosla:
## Summary

Fixes **#95**: a first prompt with no registered model dumped `typescript harness failed: ...` plus a stack.

The REPL now prints:

```
model `gpt-5.6-terra` is not registered.
  exo model register gpt-5.6-terra --secret openai
```

Other turn failures still get the usual `turn failed:` / `stream error:` prefix.

Implementation notes:

- TypeScript still throws `model is not registered: <name>`
- The host converts that exact guest message into `UnregisteredModelError`
- The TUI no longer double-prefixes with `stream error:` (it already adds that in `push_notice`)

## Test plan

- [x] `cargo test -p executor missing_model`
- [x] `cargo test -p executor guest_unregistered`
- [x] `cargo test -p exo pick_repl_model`
- [x] `cargo test -p exo missing_model_repl`
- [x] `pnpm exec oxlint exoharness/typescript/model-runtime/shared.ts`

Closes #95.


gvkhosla:
Rebased onto current `main` and resolved the CLI import conflicts from the newer sandbox types. Missing-model failures are a typed `UnregisteredModelError` rather than string sniffing.

PR #212, opened 2026-08-15
Add support for firecracker

ankrgyl:
* Run local sandboxes with firecracker, bridged through [lima](https://github.com/lima-vm/lima) on Mac OS X
* Support `fork()` (inspired by #173) so that you can warm up firecracker VMs and quickly fork them


crates/executor/src/local_sandbox.rs
```diff
@@ -328,6 +328,30 @@ impl SandboxHandle for LocalSandboxAgent {
...
+    async fn terminate_sandbox(&self, id: SandboxId) -> Result<()> {
+        let Some(local_id) = self.local_sandbox_id(&id).await? else {
+            return self.remote.terminate_sandbox(id).await;
+        };
+        self.local_agent()
+            .await?
+            .terminate_sandbox(local_id)
+            .await?;
+        self.state.sandboxes.lock().await.remove(&id);
+        self.state.detached_sandboxes.lock().await.remove(&id);
```
ankrgyl: can these be parallelized? Also what happens if one fails and bails out before the others run?

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+// guest compromises it. Bounding what root reads back keeps a hostile VMM
+// from ballooning this process's memory with endless status lines, headers,
+// or an absurd Content-Length.
+const FIRECRACKER_API_MAX_RESPONSE_BYTES: u64 = 1024 * 1024;
+const FIRECRACKER_API_TIMEOUT: Duration = Duration::from_secs(5);
+const FIRECRACKER_SNAPSHOT_CREATE_TIMEOUT: Duration = Duration::from_secs(5 * 60);
+// Firecracker forwards guest packets without filtering them. Keep special-use,
+// link-local, host-private, and cross-VM ranges closed unless explicitly admitted.
+// https://github.com/firecracker-microvm/firecracker/blob/main/docs/prod-host-setup.md#filtering-guest-egress-network-traffic
+const BLOCKED_EGRESS_CIDRS: &[&str] = &[
```
ankrgyl, line 124: i wonder if we should allow users to customize this list

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            ),
+            vcpu_count: env_parse("EXO_FIRECRACKER_VCPU_COUNT", DEFAULT_VCPU_COUNT)?,
+            memory_mib: env_parse("EXO_FIRECRACKER_MEMORY_MIB", DEFAULT_MEMORY_MIB)?,
+            image_size_gib: env_parse("EXO_FIRECRACKER_IMAGE_SIZE_GIB", DEFAULT_IMAGE_SIZE_GIB)?,
+            workspace_size_gib: env_parse(
+                "EXO_FIRECRACKER_WORKSPACE_SIZE_GIB",
+                DEFAULT_WORKSPACE_SIZE_GIB,
+            )?,
+            jailer_uid_base: env_parse("EXO_FIRECRACKER_JAILER_UID_BASE", DEFAULT_JAILER_UID_BASE)?,
+            dns_server: env_parse("EXO_FIRECRACKER_DNS_SERVER", Ipv4Addr::new(1, 1, 1, 1))?,
```
ankrgyl: these should all be clap args.

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+    clone_restore_ms: f64,
+    total_ms: f64,
+}
+
+struct PendingFirecrackerForkTiming {
+    started_at: Instant,
+    source_sync_ms: f64,
+    snapshot_create_ms: f64,
+    clone_record_ms: f64,
+}
```
ankrgyl: this code should be removed or replaced with proper tracing code

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+                    state_lock_path.display()
+                )
+            })?;
+        fs::set_permissions(&state_lock_path, Permissions::from_mode(0o600))?;
+        flock(&state_lock, FlockOperation::NonBlockingLockExclusive).with_context(|| {
+            format!(
+                "locking Firecracker state root {}; another backend may still own it",
+                config.state_root.display()
+            )
+        })?;
```
ankrgyl, line 508: should this be on a blocking thread?

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+        // Best-effort reclamation of temporaries stranded by crashed prior
+        // processes. Every publish in the state root stages under a dotted or
+        // .tmp name and renames atomically, so anything still carrying a
+        // staging name after a day is garbage that would otherwise accumulate
+        // on the root-owned volume forever.
+        super::firecracker_image::sweep_stale_image_artifacts(&config.state_root);
```
ankrgyl, line 514: this shouldn't be here? we may have solved this on `sandbox-cli`?

i dont quite think creating new sandboxes needs to involve consider old ones

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+        }
+        for directory in ["manifests", "leases"] {
+            super::firecracker_image::sweep_stale_temporaries(
+                &config.state_root.join(directory),
+                |name| name.ends_with(".tmp"),
+            );
+        }
+        config.kernel = cache_immutable_artifact(&config.state_root, "kernel", &config.kernel)?;
+        config.initramfs =
+            cache_immutable_artifact(&config.state_root, "initramfs", &config.initramfs)?;
```
ankrgyl, line 529: also worried about this blocking the async thread

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            self.shared.cleanup_machine(&entry.machine_id, true).await?;
+        }
+        let restore_started_at = Instant::now();
+        let machine = self
+            .shared
+            .ensure_machine(&request, &stable_machine_id, &spec_hash)
+            .await?;
+        let clone_restore_ms = restore_started_at.elapsed().as_secs_f64() * 1000.0;
+        self.shared.touch_machine_lease(&stable_machine_id).await?;
+        self.shared.warm_machines.lock().await.insert(
```
ankrgyl: this and the above branch look quite duplicated. and in general, i prefer `if { } else { }` rather than `if { ... return } ... `

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            request,
+            spec_hash,
+            shared: Arc::clone(&self.shared),
+            one_shot: false,
+            startup_timing,
+        }))
+    }
+}
+
+fn fork_startup_timing(
```
ankrgyl: i think we shouldn't have this

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            .shared
+            .ensure_active_machine(
+                &self.request,
+                &self.machine.record.machine_id,
+                &self.spec_hash,
+            )
+            .await?;
+        let output = GuestClient::new(Arc::clone(&self.shared), machine.vsock_path)
+            .exec(&self.request.spec, command)
+            .await;
```
ankrgyl: is it bad to have to do this for every command exec?

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            // Firecracker's clean-shutdown API is x86-only. On every architecture,
+            // sync the durable filesystem through the guest before terminating the
+            // VMM so completed writes are not stranded in the guest page cache.
+            // https://github.com/firecracker-microvm/firecracker/blob/main/docs/api_requests/actions.md#intel-and-amd-only-sendctrlaltdel
+            GuestClient::new(Arc::clone(&self.shared), self.machine.vsock_path.clone())
+                .sync_filesystem(&self.request.spec.default_workdir)
+                .await
+                .context("syncing Firecracker durable filesystem before stop")?;
+        }
+        self.shared
```
ankrgyl, line 1077: put this in an else

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            .warm_machines
+            .lock()
+            .await
+            .retain(|_, entry| entry.machine_id != self.machine.record.machine_id);
+        self.shared
```
ankrgyl, line 1082: can this be parallelized with the above?

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+                }
+                return Err(error);
+            }
+            Ok(record)
+        })
+        .await
+        .context("joining Firecracker machine allocation")?
+    }
+
+    async fn new_fork_machine_record(
```
ankrgyl: this looks mostly duplicated with the above function

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+fn parse_provider_state(value: &Value) -> Result<FirecrackerProviderState> {
+    serde_json::from_value(value.clone()).context("invalid Firecracker provider state")
+}
+
+pub(super) fn env_path(name: &str, default: &str) -> PathBuf {
+    std::env::var_os(name)
+        .filter(|value| !value.is_empty())
+        .map(PathBuf::from)
+        .unwrap_or_else(|| PathBuf::from(default))
+}
```
ankrgyl: this seems like a non-firecracker function that probably already exists. and maybe doesn't need to?

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+
+fn env_parse<T>(name: &str, default: T) -> Result<T>
+where
```
ankrgyl: this should not exist. we should be using clap and passing a struct into the firecracker code

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+{
+    match std::env::var(name) {
+        Ok(value) if !value.trim().is_empty() => value
+            .parse()
+            .map_err(|error| anyhow!("invalid {name} value {value:?}: {error}")),
+        _ => Ok(default),
+    }
+}
+
+fn validate_host(config: &FirecrackerConfig) -> Result<()> {
```
ankrgyl: should we make this async and run it on a blocking thread?

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+            }
+            Poll::Pending => Poll::Pending,
+        }
+    }
+}
+
+impl Drop for ProcessWait {
+    fn drop(&mut self) {
+        self.cleanup.take();
+    }
```
ankrgyl, line 3908: i feel like this should be from an existing library...

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+}
+
+fn record_launch_timing(machine_id: &str, step: &str, duration: Duration) {
```
ankrgyl: this looks like debug code that we should remove

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+        }
+        tokio::time::sleep(Duration::from_millis(1)).await;
+    }
+    bail!(
+        "restored Firecracker guest agent did not become ready: {}",
+        tail_file(&stderr_path)?
+    )
+}
+
+fn wait_for_guest_blocking(
```
ankrgyl, line 3959: this is a good naming convention that we should use for all the other synchronous functions that might block (i've pointed this out a bunch above)

crates/exoharness/src/sandbox_provider/firecracker.rs
```diff
@@ -0,0 +1,4021 @@
...
+        }
+    };
+    if touched {
+        shared.touch_machine_lease(machine_id).await?;
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
ankrgyl: can you move these into a separate file, like firecracker_tests.rs

crates/exoharness/src/sandbox_provider/firecracker_image.rs
```diff
@@ -0,0 +1,1724 @@
...
+            )
+        }),
+    }
+}
+
+struct LimitedAsyncWriter<W> {
+    inner: W,
+    remaining: u64,
+}
+
```
ankrgyl, line 586: we should be able to use a library for this?

crates/exoharness/src/sandbox_provider/firecracker_image.rs
```diff
@@ -0,0 +1,1724 @@
...
+    Ok(Box::new(BoundedReader {
+        inner: BufReader::new(file),
+        remaining: decompressed_budget,
+    }))
+}
+
+struct BoundedReader<R> {
+    inner: R,
+    remaining: u64,
+}
```
ankrgyl, line 958: there should be a library for this

ankrgyl: Kept this adapter. `std::io::Take` turns the size limit into ordinary EOF, so it cannot distinguish a stream that ends exactly at the budget from a decompression bomb that has more output. The one-byte probe at the boundary is the security property this wrapper adds.

crates/exoharness/src/sandbox_provider/firecracker_image.rs
ankrgyl: Kept this adapter. Tokio has bounded readers, but no bounded `AsyncWrite` adapter with the needed behavior: reject data beyond the OCI descriptor size before accepting it. This protects the host disk while the untrusted registry stream is being written; the digest check afterward handles content integrity.

crates/exoharness/src/sandbox_provider/firecracker.rs
ankrgyl: Kept cleanup unconditional. This condition controls only whether the durable workspace needs a guest-side sync before shutdown. Putting cleanup in an `else` would skip VM cleanup whenever that sync path runs.

crates/exoharness/src/sandbox_provider/firecracker.rs
ankrgyl: Kept the denylist as the secure baseline. Operators can already admit specific private or special-use ranges with the repeatable `--firecracker-allowed-egress-cidr` option. Replacing the baseline wholesale would make it too easy to expose host-local and cross-VM ranges.

crates/exoharness/src/sandbox_provider/firecracker.rs
ankrgyl: Kept the small wrapper. It attaches Firecracker-specific cleanup to the generic process-bridge wait future and guarantees that cleanup runs exactly once whether the wait completes or the caller drops it. Tokio/futures do not provide that ownership semantic directly, so a library would still require this adapter.

crates/exoharness/src/sandbox_provider/firecracker.rs
ankrgyl: Did not parallelize these. The warm-map eviction must happen before destructive process, network, and filesystem cleanup so no lookup can return a machine while it is being torn down. Both operations are serialized by the lifecycle lock; parallel execution would introduce a race without a useful latency win.

crates/exoharness/src/sandbox_provider/firecracker.rs
ankrgyl: Kept this. It does not reap VMs or inspect sandbox lifecycle; it only removes unpublished temporary files older than a day after the backend has acquired the exclusive state-root lock. Published state is atomically renamed out of these names, so this is crash-debris GC rather than sandbox cleanup.

Nandinitalwar (approved):
lgtm! non-blocking but could add an early /dev/kvm check for unsupported setups on m1/m2 macs

ankrgyl:
> lgtm! non-blocking but could add an early /dev/kvm check for unsupported setups on m1/m2 macs

Good idea just added

Nandinitalwar (approved):

akrentsel (approved):

akrentsel:
a few small fixes noticed in reviewing this PR, merged one in #220 and a few others coming in #221, but otherwise lgtm (and working locally having played with it). Thanks!

PR #213, opened 2026-08-16
Add Pi coding-agent harness preset

gvkhosla:
## Summary

Adds a Pi coding-agent harness using the same Exoharness integration pattern as the existing Codex, Claude Code, and Cursor harnesses.

- adds a `--harness pi` TypeScript preset with a default sandbox image and networking
- runs Pi inside the configured Exoharness sandbox
- treats Exoharness events as canonical history and reconstructs context on each turn
- projects Pi text, tool calls/results, retries, compaction, raw messages, per-call usage/timing, and completion into the standard event and trace paths
- bounds history replay and worker lifetime, scopes JSONL events to turn IDs, and forwards agent output/round limits
- adds the sandbox image, CLI/unit coverage, live harness E2E entry, and matching documentation

## Implementation

The implementation follows the existing coding-agent harness structure rather than introducing a new executor path:

- `pi-harness.ts` uses `traceExecutorTurn`, `tracedUnderParent`, `resolveLlmBinding`, `materializePriorConversationMessages`, `WarmJsonlSandboxWorker`, and `WarmResourceCache`, matching the Cursor harness shape.
- A JSONL SDK worker runs inside the sandbox, so Pi's native read/write/edit/bash tools stay within the Exoharness sandbox boundary.
- Each request uses `SessionManager.inMemory()`, similar to Claude Code's `persistSession: false`: Exoharness remains the only durable conversation history.
- The worker uses an explicit resource loader so it does not implicitly load host or project Pi extensions, skills, prompts, settings, context files, or sessions.
- Model secrets and base URLs come from the existing Exoharness LLM binding path. No binding schema or Exoharness API changes are required.

This is additive and does not change Codex, Claude Code, Cursor, or the underlying Exoharness/executor boundary.

## Initial scope

- Model IDs use Pi's `provider/model` form and Pi's built-in model catalog.
- Provider credentials enter the sandbox because Pi makes model requests there, matching the current coding-agent sandbox pattern. The worker removes the credential from its environment before Pi tools run, so tool subprocesses do not inherit it.
- Pi extensions and skills remain disabled by default; exposing additional resources can be considered separately.

## Validation

- `pnpm lint`
- `pnpm typecheck`
- `pnpm test` — 166 passing
- `pnpm format:check`
- `cargo test --workspace --all-targets`
- `pnpm docs:build`
- Pi sandbox image built successfully with Docker
- sandbox worker exercised against a mocked Anthropic SSE endpoint, including tools, warm-worker request isolation, limits, usage/timing, and credential non-inheritance
- full two-turn `exo -> Pi sandbox -> mocked provider` run verified canonical Exoharness history replay

The live E2E entry is available as:

```bash
pnpm e2e:agent-harnesses --only pi --build-images
```

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -57,7 +57,7 @@ use tui::run_chat_repl;
 struct Cli {
     #[arg(long, global = true, default_value = ".exo")]
     root: PathBuf,
-    /// Executor runtime: basic, rlm, typescript, codex, claude-code, cursor, or a TypeScript module path.
+    /// Executor runtime: basic, rlm, typescript, codex, claude-code, cursor, pi, or a TypeScript module path.
     #[arg(long, global = true, value_name = "HARNESS")]
     harness: Option<HarnessSelection>,
     #[arg(long, global = true, value_enum, env = "EXO_SECRET_BACKEND")]
@@ -124,6 +124,7 @@ enum TypeScriptHarnessPreset {
     Codex,
     ClaudeCode,
     Cursor,
+    Pi,
 }
 
 impl HarnessSelection {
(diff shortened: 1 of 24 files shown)


gvkhosla:
Closing in favor of the in-tree pi-harness work on origin/pi-harness (minimal Pi harness, cold per turn). This PR is stale against current main. Happy to open a follow-up against that branch if anything here is still useful.

PR #216, opened 2026-08-17
Rework sandbox support in the CLI

ankrgyl:
Make it much easier to launch sandboxes from the CLI


crates/cli/src/main.rs
```diff
@@ -668,6 +635,119 @@ enum ConversationSandboxCommands {
     },
 }
 
+#[derive(Debug, Subcommand)]
```
akrentsel, line 638: thanks these are great

PR #223, opened 2026-08-21
Reflect on graded Harbor trials

akrentsel:
With `--reflection`, each trial forks its attached container before Harbor grades it; the plugin then has Exo review the grade inside the fork and terminates it. Adds `exo sandbox select`/`deselect`/`fork` and lets forks and snapshots read attached Docker containers.


crates/cli/src/main.rs
```diff
@@ -2069,57 +2141,6 @@ async fn main() -> Result<()> {
                 }
             },
             ConversationCommands::Sandbox { command } => match command {
-                ConversationSandboxCommands::Attach {
```
akrentsel, line 2072: Goal of attach is to add a handle to an already running container into the Exoharness – doesn't need to be tied to a conversation.

PR #227, opened 2026-08-22
fix: resolve DEFERRED_SCRIPT path so rebuild_and_restart_exo can spawn the deferred runner (#214)

### LINKED ISSUE #214: rebuild_and_restart_exo: DEFERRED_SCRIPT path is wrong (ENOENT), self-update stays "queued"
## Summary

`rebuild_and_restart_exo` resolves the deferred-restart helper to a path that does not exist, so the spawned process fails with `ENOENT` and the queued self-update never completes.

## Location

`exo/tools/guardian-tools.ts` (at current `main`, `5bc77ce`):

```ts
const GUARDIAN_SCRIPT = new URL(
  "../scripts/exo-service-guardian",          // correct
  import.meta.url,
).pathname;
const DEFERRED_SCRIPT = new URL(
  "./scripts/deferred-rebuild-and-restart",   // resolves under exo/tools/scripts/
  import.meta.url,
).pathname;
```

From `exo/tools/guardian-tools.ts`, `./scripts/deferred-rebuild-and-restart` resolves to `exo/tools/scripts/deferred-rebuild-and-restart`, which does not exist. The actual script is at `exo/scripts/deferred-rebuild-and-restart`. The sibling `GUARDIAN_SCRIPT` a few lines above correctly uses `../scripts/`.

It looks like the refactor in #187 corrected `GUARDIAN_SCRIPT`'s relative path but missed `DEFERRED_SCRIPT`.

## Effect

`runGuardianDeferredWithOutcome` spawns `DEFERRED_SCRIPT`. With the wrong path the child fails to spawn (`ENOENT`), so the deferred runner never executes `restart-all --build` and never finalizes the outcome. The durable update record written by `queueRebuildAndRestart` stays `status: "queued"` indefinitely, and `list_conversation_events` never sees a completion event for the rebuild.

## Suggested fix

```ts
const DEFERRED_SCRIPT = new URL(
  "../scripts/deferred-rebuild-and-restart",
  import.meta.url,
).pathname;
```

## Note

`exo/tools/guardian-tools.test.ts` covers registration, schema, and reason parsing, but not the spawn path, so this is not caught by the current tests.

Prithvi1994:
Closes #214

## Problem
The `DEFERRED_SCRIPT` path in `exo/tools/guardian-tools.ts` was set to `"./scripts/deferred-rebuild-and-restart"` relative to `guardian-tools.ts`. This resolves to `exo/tools/scripts/deferred-rebuild-and-restart`, which does not exist. As a result, the deferred rebuild runner could never be spawned, causing `rebuild_and_restart_exo` to stay in the `"queued"` state forever with an `ENOENT` error.

## Fix
Changed the relative URL to `"../scripts/deferred-rebuild-and-restart"` so it resolves correctly from `exo/tools/` up to `exo/scripts/` where the actual bash script lives.

## Verification
- The path now points at the existing file: `exo/scripts/deferred-rebuild-and-restart`.
- The TypeScript compiles (no new errors).
- The fix is minimal and surgical—one line changed.

Closes #214.

### DIFF
--- a/exo/tools/guardian-tools.ts
+++ b/exo/tools/guardian-tools.ts
@@ -22,7 +22,7 @@ const GUARDIAN_SCRIPT = new URL(
   import.meta.url,
 ).pathname;
 const DEFERRED_SCRIPT = new URL(
-  "./scripts/deferred-rebuild-and-restart",
+  "../scripts/deferred-rebuild-and-restart",
   import.meta.url,
 ).pathname;
 const ROOT_DIR = new URL("../..", import.meta.url).pathname;


Prithvi1994:

PR #228, opened 2026-08-23
Fix agent sandbox creation and process bridge failures

derekdeming:
## Summary

- serialize agent-scoped sandbox initialization so concurrent conversations cannot create multiple supposedly shared sandboxes
- reject unsuccessful process-bridge responses for output polling, stdin writes, and stdin closure instead of spinning or silently accepting the failure
- add one regression test covering process-bridge error termination

## Validation

- `cargo test -p exoharness --features basic-backend -p executor`
- `oxlint --deny-warnings .`
- `tsgo --noEmit -p tsconfig.json`
- `vitest run`

### DIFF
--- a/crates/executor/src/agent_sandbox.rs
+++ b/crates/executor/src/agent_sandbox.rs
@@ -1,8 +1,12 @@
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, OnceLock};
+
 use exoharness::{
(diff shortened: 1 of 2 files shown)


derekdeming:
Personally noticed two issues while playing around with the harness. Exo could accidentally create 2 sandboxes for the same agent when conversations started simultaneously. Then it could also ignore a failed response from aws agentcore or firecracker and poll forever. The PR ensures each agent gets only 1 shared sandbox and stops failed remote commands immediately

PR #233, opened 2026-08-24
fix: correct DEFERRED_SCRIPT path in guardian-tools

### LINKED ISSUE #214: rebuild_and_restart_exo: DEFERRED_SCRIPT path is wrong (ENOENT), self-update stays "queued"

Prithvi1994:
Fixes #214

`DEFERRED_SCRIPT` in `exo/tools/guardian-tools.ts` resolved `./scripts/deferred-rebuild-and-restart` to the nonexistent `exo/tools/scripts/deferred-rebuild-and-restart`, so the spawned deferred runner failed with ENOENT and queued self-updates stayed `status: "queued"` forever. The script actually lives at `exo/scripts/deferred-rebuild-and-restart` — use `../scripts/`, matching the sibling `GUARDIAN_SCRIPT` (whose path was corrected in the same refactor that missed this one).

One-line change; verified the target path exists on `main`.

### DIFF
--- a/exo/tools/guardian-tools.ts
+++ b/exo/tools/guardian-tools.ts
@@ -22,7 +22,7 @@ const GUARDIAN_SCRIPT = new URL(
   import.meta.url,
 ).pathname;
 const DEFERRED_SCRIPT = new URL(
-  "./scripts/deferred-rebuild-and-restart",
+  "../scripts/deferred-rebuild-and-restart",
   import.meta.url,
 ).pathname;
 const ROOT_DIR = new URL("../..", import.meta.url).pathname;


CakeCrusher:
@ankrgyl plz merge

ankrgyl (approved):

ankrgyl:
thanks for contributing!

Issue #207, opened 2026-08-11
Attention Scheduler Proposal


ad1269:
## Problem / Current State

[`send_conversation_wakeup`](https://github.com/exoharness/exo/blob/5bc77ce7c7a2921794083d58c926cf721c14bf8a/crates/executor/src/conversation_wakeup.rs#L18) is the function through which anything outside a turn reaches a conversation. It works by acquiring a lock file, and whichever process wins starts the next turn and holds the lock for that turn's entire duration. This has the following downsides:

* Whichever polling mechanism happens to see the lock first wins, losing arrival order
* The lock has no heartbeat, so a turn running past 30 minutes looks identical to a crash. A waiting producer reclaims the lock and starts a second concurrent turn. The first turn ending then reclaims the lock for the second turn, allowing a third turn to start. Everything after this point is turns interleaving and writing to the event log.
* The scheduler correctly handles the ordering of various events, but that is lost by this layer, and so we can't make any ordering guarantees on when things actually get into the agent
* Nothing is durable here, and producers block until their data is picked up. A crash loses the data held in the producer.

### Agent Inbox

Instead of this lock file mechanism, we should have a durable Agent Inbox. Producers should durably append items to this inbox, and a dispatcher should grab items from it and inject items into the conversation.

```rust
struct InboxItem {
    item_id: Uuid7,                // identity of the logical fact, not a delivery attempt
    conversation_id: String,
    producer: ProducerRef,         // { kind: scheduler | adapter | agent, id }
    /// Per-OCCURRENCE dedupe: (task_id, slot) for fires; platform message id
    /// qualified by adapter and target for inbound — the same identities
    /// ingest already dedupes on, lifted so they travel with the item.
    dedupe_key: String,
    attention: Attention,          // the one cross-producer field, below
    native: serde_json::Value,     // producer-native semantics, opaque to the runtime
    prompt: String,                // rendered payload
    artifacts: Vec<ArtifactRef>,
    appended_at_ms: u64,
}
```

Inbox Items should declare an attention class so the dispatcher can decide how to handle that item. These 3 enum values are essentially the only things a dispatcher can do with a pending Inbox Item, which is why it's a closed set of actions rather than a priority rank.

```rust
enum Attention {
    /// May enter a RUNNING turn at the next round boundary.
    Interrupt,
    /// Opens a turn when the conversation is quiescent. FIFO among peers.
    Wake,
    /// Held for coalescing; flushed with the next turn or when max_wait
[...]

PR #234, opened 2026-08-24
feat: durable ordered Agent Inbox for conversation wakeups (#207)

Prithvi1994:
Part of #207 — durable inbox primitive (first pass)

## Problem
`send_conversation_wakeup` used a lock file in `/tmp`: producers raced on `create_new`, arrival order was lost, and a 30-minute mtime heuristic let any waiting producer steal a lock from a legitimately slow turn — producing concurrent turns interleaving writes into one event log. Nothing was durable; a crash lost the prompt.

## This PR
Adds `crates/executor/src/inbox.rs`, a durable per-conversation inbox:

- **enqueue()** writes `<root>/<conversation_id>/<uuid7>.json` via temp-file + rename. Producers return once durable.
- **drain()** lists pending items and sorts by `Uuid7` — time-ordered IDs make FIFO delivery free, restoring arrival-order guarantees the lock race destroyed.
- **ack()** renames to `.json.done` after successful send → at-least-once semantics: crash between send and ack redelivers on next drain.

`send_conversation_wakeup_content` now enqueues first, then under the existing in-process per-conversation async mutex drains all pending items in order and injects each via `conversation.send`.

The racy `WakeupFileLock` + 30-minute stale-lock steal is deleted from the delivery path entirely.

## What this fixes (vs issue #207)
| Issue complaint | Status |
|---|---|
| Lost arrival ordering | ✅ fixed within/between processes sharing an inbox root (UUIDv7 sort) |
| No durability / producers block | ✅ enqueue returns after durable write |
| Ordering info discarded | ✅ preserved by UUIDv7 keys |
| 30-min lock steal → concurrent turns | ✅ removed from delivery path; turn mutual exclusion is now the in-process mutex |

## Deliberately out of scope (follow-up)
This PR is the **storage primitive half** of #207. Not included here:

- `producer` / `dedupe_key` / `attention` fields on `InboxItem` — the idempotent-append + attention-scheduler half
- Dispatcher `decide(state, pending, now_ms)` and the `Attention` classes (`Interrupt` / `Wake` / `Batch`)
- `retract()` / cancellation of undrained items
- ack at **turn commit** + `OfferDisposition` settlement
- Cross-process per-conversation dispatch lease (depends on #113)

See the scoping comment for the full gap table.

## Verification
`cargo test -p executor`: 120 passed (6 new inbox tests + reworked lock test), 0 failed.

Configurable root via `EXO_INBOX_DIR` so deployments can co-locate it with durable storage.

### DIFF
--- a/crates/executor/src/conversation_wakeup.rs
+++ b/crates/executor/src/conversation_wakeup.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex, OnceLock};
-use std::time::{Duration, SystemTime};
 
 use anyhow::Context;
 use exoharness::Result;
@@ -9,7 +8,19 @@ use lingua::Message;
 use lingua::universal::UserContent;
 use tokio::sync::Mutex as AsyncMutex;
 
+use crate::inbox::{Inbox, InboxItem};
 use crate::{HarnessConversation, SendRequest, SendResult};
+
+/// Default on-disk location of the durable conversation inbox.
+///
+/// Overridable via `EXO_INBOX_DIR` so deployments can place it on a
+/// volume with the same durability guarantees as the event log.
+pub fn default_inbox() -> Inbox {
+    let root = std::env::var_os("EXO_INBOX_DIR")
+        .map(PathBuf::from)
+        .unwrap_or_else(|| std::env::temp_dir().join("exo-inbox"));
+    Inbox::new(root)
+}
 
 pub async fn send_conversation_wakeup(
     conversation: &dyn HarnessConversation,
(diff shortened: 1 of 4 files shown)


Prithvi1994:
## Scoping note vs #207

This PR lands the durable inbox **storage primitive**. Flagging explicitly what is deferred so the remaining work on #207 stays tracked rather than implied as done.

### Delivered here

| #207 requirement | Status |
|---|---|
| Durable append | ✅ temp-write + rename |
| Arrival order | ✅ UUIDv7 sort (FIFO) |
| Producer crash safety | ✅ enqueue returns post-durable; unacked items redelivered |
| Lock-file race + 30-min stale steal | ✅ removed from delivery path |

### Deferred (follow-up PRs)

| #207 requirement | Why deferred |
|---|---|
| `producer`, `dedupe_key`, `attention`, `native`, `artifacts`, `appended_at_ms` on `InboxItem` | Idempotent-append half; needs conditional-put from #113 |
| Idempotent append per `(conversation, dedupe_key)` | Scheduler fires / adapter inbound retries would currently duplicate |
| `retract()` → `Retracted or AlreadyDrained` (cancel undrained items) | Pairs with the attention dispatcher |
| `Attention` enum (`Interrupt` / `Wake` / `Batch`) | The scheduler decision surface; issue author agreed to defer `Preempt` |
| Dispatcher `decide(state, pending, now_ms) -> Action` | Needs coordinator `list_pending` + dedupe key surviving turn end |
| ack **only at turn commit** + `OfferDisposition` (`Consumed` / `Released` / `Remainder`) | Current ack is right after each send; no give-back on an uncommitted turn |
| Coalescing N pending items into one turn | Current behavior opens N sequential turns — one per item |
| Cross-process per-conversation dispatch lease | The reason the lock file existed; depends on #113 |

### Two behavioural notes to settle in the follow-up

1. `send_conversation_wakeup_content` opens a turn **per pending item**, whereas #207's dispatcher table specifies **one** turn draining all pending items FIFO.
2. Ack currently follows each successful `send`, not the turn's commit — at-least-once holds, but there is no "a turn that ends without committing gives back everything it held" path yet.

Happy to take the `Attention` + dispatcher follow-up once #113 lands.

PR #236, opened 2026-08-25
Keep ExoWorker prompts inside the model context window + add grok 4.6 context window

Bautista1999:
## Summary
- Cap tool results in the prompt (32K for the latest result, 4K for older ones) and compress older turns before each model call, pinning developer instructions, the first user message, and recent tool pairs, with a compression marker so later turns stay stable.
- Resolve per-model context windows (including Grok 4.6 at 500K) and retry on overflow: force-compress, then strip vision image parts that providers bill far above our estimate.
- Tighten vision attaches to 1 image / 120KB per round, skip path-only `previewPresentation` results, and drop host-only `_visionSlides` / `_previewFiles` batches from tool JSON after extract.

## Test plan
- [ ] `pnpm test examples/exo-worker` (context window, compression selection/markers, vision strip, tool-result caps, path-only preview)
- [ ] Long thread that would exceed the window: compression runs before the model call; later turns still see the marker
- [ ] Force a context-window 400: retry strips remaining images and completes
- [ ] `previewPresentation` with disk paths only: no vision attach; nested slide previews attach at most one image
- [ ] `/manage` or exo `--model grok-4.6`: compression budgets use 500K, not the 200K default


akrentsel (approved):
you'll need to fix the merge conflcits! but otherwise lgtm

PR #241, opened 2026-08-28
add support for pi harness

akrentsel:
Adding pi harness integration as an executor option. This will allow running benchmarks against Pi.

This integration runs Pi inside the sandbox – we use the dockerfile that already loads in the Pi process – and populates the API key from the secret store.

We keep the exoharness state as the source of truth, and also record cost figures coming from Pi.


ankrgyl (approved):
lgtm

PR #243, opened 2026-08-29
Generalize snapshot formats

ankrgyl:
Instead of a fixed enum of snapshot formats, use a string, so that each provider can perform its own snapshot (including arbitrary trait implementations which are not in the main repo)

### DIFF
--- a/exoharness/docs/sandbox-snapshots.md
+++ b/exoharness/docs/sandbox-snapshots.md
@@ -1,13 +1,13 @@
 # Sandbox Snapshots
 
-Status: implemented for the Docker, E2B, Sprites, and Daytona sandbox
-backends; stubs in place for the others. Daytona can additionally restore a
-Docker snapshot (the cross-provider "teleport" bridge).
+Status: implemented for Docker, Daytona, E2B, Sprites, SmolVM, and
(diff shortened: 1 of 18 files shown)


exoharness/docs/sandbox-snapshots.md
```diff
@@ -199,17 +225,18 @@ cargo test -p exo --test teleport_docker_to_daytona -- --ignored --nocapture
...
 2. Implement `ManagedSandboxHandle::snapshot` for that backend's handle
    type, producing the appropriate `SnapshotPayload`. The Docker version in
    `docker_snapshot_container` is the template — three CLI calls and a
    `Bytes` capture.
-3. Implement `ManagedSandboxBackend::acquire_from_snapshot` to consume the
-   same `kind`, including the safety check that the payload's `kind`
-   matches what the backend understands. The Docker version is the
-   template here too — load the bytes, get the loaded image reference,
-   swap `request.spec.image`, evict + recreate the warm container.
+3. Return every supported identifier from
```
akrentsel, line 235: yeah this makes more sense, thanks for the change

akrentsel:
Looks great, thanks

PR #261, opened 2026-09-04
feat: add JSON-RPC agent server over stdio

kumanday:
## Summary

Adds `exo agent-server`, a long-lived JSON-RPC 2.0 server that exposes Exo’s executor harness over newline-delimited stdin/stdout.

This allows external processes to resolve persisted agents and conversations, start model/tool turns, consume typed streaming events, and correlate results with durable Exoharness identities without embedding the executor.

## What changed

- Add the `exo agent-server` CLI command.
- Add protocol initialization and capability discovery.
- Support:
  - `agent/list`
  - `agent/get`
  - `conversation/list`
  - `conversation/get`
  - `conversation/create`
  - `turn/start`
- Stream typed `turn/started`, `turn/event`, `turn/completed`, and `turn/failed` notifications.
- Return process-local operation IDs and native session, turn, and event IDs.
- Enforce one active turn per conversation while allowing different conversations to run concurrently.
- Reject agents whose persisted harness does not match the server runtime.
- Keep stdout protocol-only and send diagnostics to stderr.

## Protocol and lifecycle behavior

- Uses JSON-RPC 2.0 with JSONL framing.
- Requires initialization with protocol version 1.
- Preserves `null` request IDs and returns parse or invalid-request errors without terminating the server.
- Rejects frames larger than 8 MiB and recovers from malformed UTF-8 or JSON.
- Owns executor stream tasks and cancels/joins them on stdin EOF or output failure.
- Preserves an already-finalized completion during shutdown.
- Contains executor-task panics to the affected operation.
- Releases the active-conversation guard before terminal notifications become observable.
- Invalidates and terminates failed or cancelled TypeScript runners before queued turns may reuse them.
- Returns bounded client-safe executor errors while retaining full diagnostics in default stderr logging.

`turn/cancel` remains intentionally unsupported and is advertised as unavailable. The server can cancel owned turns during shutdown, but client-requested cancellation will wait until operation lookup, race handling, and terminal notification semantics are fully defined.

## Documentation

Adds `docs/agent-server.md` and README discovery for the new command and protocol.

## Verification

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test --workspace --all-targets` — 346 passed, 0 failed
- `pnpm format:check`
- `pnpm check` — lint and typecheck passed; 146 tests passed

Credential-dependent and live integration tests remain ignored as before.

### DIFF
--- a/crates/executor/src/agent_server/server.rs
+++ b/crates/executor/src/agent_server/server.rs
@@ -1,0 +1,816 @@
+use std::collections::HashSet;
+use std::sync::Arc;
+
+use anyhow::{Context, Result, anyhow};
+use exoharness::{AgentId, ConversationId, Uuid7};
+use serde::Serialize;
+use serde::de::DeserializeOwned;
+use serde_json::value::RawValue;
+use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader, BufWriter};
+use tokio::sync::Mutex;
+use tokio::task::{JoinError, JoinSet};
+use tokio_stream::StreamExt;
+use tokio_util::sync::CancellationToken;
+
+use crate::{
+    AgentHarnessKind, CreateConversationRequest, ExecutionStreamEvent, ExecutionStreamHandle,
+    Harness, HarnessConversation, SendRequest,
+};
+
+use super::AGENT_SERVER_TRACING_TARGET;
+use super::protocol::{
+    AgentGetParams, AgentGetResult, AgentListResult, ConversationCreateParams,
+    ConversationCreateResult, ConversationGetParams, ConversationGetResult, ConversationListParams,
+    ConversationListResult, EmptyParams, InitializeParams, InitializeResult, OperationError,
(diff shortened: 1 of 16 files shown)


crates/executor/src/agent_server/server.rs
```diff
@@ -0,0 +1,816 @@
...
+            .get_agent(agent_ref)
+            .await
+            .map_err(RpcError::executor)?
+            .ok_or_else(|| agent_not_found(agent_ref))
+    }
+
+    async fn start_turn(
+        &self,
+        params: TurnStartParams,
+    ) -> std::result::Result<(ResponseResult, Option<TurnOperation>), RpcError> {
```
ankrgyl, line 299: what does this actually do? Im probably missing something obvious but it seems like it just inserts the turn into a map?

kumanday: This function prepares the turn and marks the conversation as busy. After sending the client an acceptance response, `serve()` starts it in the background through `conversation.send_stream()` and streams the results back. `prepare_turn()` would be a clearer name.

PR #262, opened 2026-09-04
Reconnect ExoChat after WebSocket drops

alanagoyal:
## Summary

- reconnect the ExoChat browser client after dropped relay WebSockets with exponential backoff from 1 to 30 seconds
- preserve unsent drafts while disconnected and never queue or automatically replay messages
- surface reconnecting, restored, and replaced states in the UI
- stop retrying when the relay replaces an older connection for the same role
- cover retry timing, reset behavior, terminal replacement, session expiry, stale events, shutdown, and send behavior with focused unit tests

[download.webm](https://github.com/user-attachments/assets/e4ed25bf-c3e9-4797-b0eb-13017ae58288)

## Testing

- `pnpm check` (15 files, 154 tests)
- `pnpm format:check`
- `pnpm website:build`
- `cargo test --workspace --all-targets`
- manual local relay test: typed an unsent draft, stopped and restarted Wrangler, verified automatic recovery and draft preservation, sent the message once, then opened a newer same-role tab and verified the old tab entered the terminal replaced state

### DIFF
--- a/website/src/chat.jsx
+++ b/website/src/chat.jsx
@@ -38,6 +38,7 @@ import {
   MessageScrollerViewport,
 } from "@/components/ui/message-scroller";
 import { Textarea } from "@/components/ui/textarea";
+import { createReconnectingWebSocket } from "@/reconnecting-websocket";
 import "./chat.css";
 
 function ChatApp() {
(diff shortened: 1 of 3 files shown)


ankrgyl (approved):
seems fine

PR #267, opened 2026-09-06
Separate sandbox identity from owner scope and trace Firecracker lifecycle

ankrgyl:
Sandbox requests now carry a required `sandbox_id` and an optional `SandboxScope` (`Agent` or `Thread`). Scope contains only the owner ID. Raw allocation needs no scope; harness requests retain their owner context. Provider lookup, warm caches, lifecycle locks, and machine names use the sandbox ID independently of scope.

Firecracker operations gain standard tracing spans, including propagation across blocking launch and snapshot tasks. The optional `ManagedSandboxHandle::is_running` observation checks VMM process exit through a host pidfd without contacting the guest; unsupported backends report no observation. Request arguments and payloads are excluded from spans.

Validation: workspace tests and workspace Clippy with warnings denied pass locally, along with the focused Firecracker tests. Coverage checks scope-free request serialization, required sandbox identity, harness routing, and provider naming independent of scope.

### DIFF
--- a/crates/exoharness/src/sandbox.rs
+++ b/crates/exoharness/src/sandbox.rs
@@ -24,37 +24,7 @@ use uuid::Uuid;
 
 use crate::{DurableFileSystem, SandboxAttachment};
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub enum SandboxKey {
-    StandaloneSandbox {
-        sandbox_id: String,
-    },
-    AgentSandbox {
-        agent_id: String,
-        sandbox_id: String,
-    },
-    ConversationSandbox {
-        #[serde(alias = "conversation_id")]
-        thread_id: String,
-        sandbox_id: String,
-    },
-}
-
-impl fmt::Display for SandboxKey {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            Self::StandaloneSandbox { sandbox_id } => write!(f, "sandbox:{sandbox_id}"),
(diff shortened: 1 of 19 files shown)


crates/exoharness/src/sandbox.rs
```diff
@@ -1277,8 +1276,8 @@ pub(crate) fn stable_fnv1a_hex(input: &str) -> String {
 }
 
 async fn touch_warm_sandbox(
-    warm_sandboxes: &Arc<Mutex<HashMap<SandboxKey, WarmSandboxEntry>>>,
-    key: &SandboxKey,
+    warm_sandboxes: &Arc<Mutex<HashMap<String, WarmSandboxEntry>>>,
+    key: &str,
```
ankrgyl: shouldn't this be the id? and it's a uuid? and the key can be a uuid (or whatever the exo id type is)?

crates/exoharness/src/sandbox.rs
```diff
@@ -1371,7 +1373,7 @@ async fn find_running_warm_sandbox(
 async fn find_sandbox_containers_for_key(
     container_bin: &Path,
     cli: ContainerCliFlavor,
-    key: &SandboxKey,
+    key: &str,
```
ankrgyl: same key thing

crates/exoharness/src/sandbox.rs
```diff
@@ -1371,7 +1373,7 @@ async fn find_running_warm_sandbox(
 async fn find_sandbox_containers_for_key(
```
ankrgyl: also should we rename this to id?

PR #277, opened 2026-09-08
scheduler: don't resurrect a task deleted mid-run (#276)

### LINKED ISSUE #276: Scheduler resurrects a task deleted while it was running
## Summary

`run_task` writes the task record back after firing without checking that the task still exists. A delete that lands mid-run is silently undone, and the task keeps firing.

`crates/executor/src/scheduler_runtime.rs`:

```rust
task.resume_after_fires(&plan, now_ms());
store.put_task(&task).await?;   // recreates a task deleted during the run
```

## Why it matters

A fire runs a full model turn, so the window is long — seconds to minutes. `delete_scheduled_task` and `cancel_scheduled_task` are model-facing tools, so the natural way to stop a task is from inside a turn, which is exactly when this window is open. The delete appears to succeed and the task returns.

There is also no operator escape hatch: `exo --help` exposes no scheduler subcommand, so stopping a runaway task otherwise means editing `.exo/scheduled-tasks/tasks/` by hand.

<details>
<summary>How I hit this</summary>

An `@every 1m` task bound to a long-lived conversation. Each fire ran a model turn whose context was the whole conversation history, so turns were slow and the delete window was wide. Combined with a task that reappeared after deletion, the loop kept consuming credits with no obvious way to stop it from inside the agent.

</details>

## Fix

Check the task still exists before putting it back. PR to follow.

This narrows the window rather than closing it — there is still a gap between the read and the write. A stronger fix would make the write conditional at the store layer (compare-and-swap or a delete tombstone). Happy to go that route instead if preferred; #232 is touching `scheduler_store.rs` and may be the better home for it.

---
Written by Claude, prompted and reviewed by me.

CakeCrusher:
Fixes #276.

A fire runs a model turn, so a task can be deleted while a run holds a copy of its record. `run_task` then wrote that stale copy back, and because `put_task` created the file when it was missing, the delete was undone and the task kept firing. `delete_scheduled_task` and `cancel_scheduled_task` are model-facing tools, so a delete naturally arrives during a run — that is how this was hit.

`create_task` is the only creator, so `put_task` becomes replace-only: a no-op once the task is gone. The existence check and the write happen under an exclusive lock on the tasks directory, taken by `delete_task` as well, so a delete cannot land between them. The lock is held around the file operations only, never across a run, and the OS releases it if a holder dies (`std::fs::File::lock`, stable since 1.89 — no new dependency).

Deletion is therefore terminal: an in-flight fire still finishes, and the task is simply never scheduled again.

The whole change is in `scheduler_store.rs`. **No call sites change** — `run_task` is untouched.

Includes a test that reproduces the original failure: a run holds a record, a delete lands mid-run, and the stale write-back must not recreate the task. It fails if `put_task` is allowed to create.

<details>
<summary>Related, if useful</summary>

The doc comment on `claim_due_tasks` notes its read-lease-write is unconditional pending "the conditional puts in upstream PR #113". Replace-only `put_task` also covers that call site.

Worth flagging: #113 as it stands keeps the unconditional `put_task` at the end of `run_task`, so this bug survives it — and #113 removes `claim_due_tasks` entirely. Happy to rebase onto #113 or #232 in whatever order suits.

</details>

`cargo test -p executor` (115 passed), `cargo clippy --all-targets -D warnings`, `cargo fmt`, and `cargo check --workspace --all-targets` all clean.

https://claude.ai/code/session_01QhK4EoQJsbjEaqGR3VJ7g5

### DIFF
--- a/crates/executor/src/scheduler_store.rs
+++ b/crates/executor/src/scheduler_store.rs
@@ -26,7 +26,7 @@ impl SchedulerStore {
 
     pub async fn create_task(&self, request: NewScheduledTask) -> Result<ScheduledTaskRecord> {
         let task = ScheduledTaskRecord::new(request, now_ms())?;
-        self.put_task(&task).await?;
+        self.write_task(&task, Write::Create).await?;
         Ok(task)
     }
 
@@ -114,12 +114,31 @@ impl SchedulerStore {
         }
     }
 
+    /// Updates an existing task record.
+    ///
+    /// A fire runs a model turn, so a task can be deleted while a run holds a
+    /// copy of its record. Writing that copy back must not recreate the task,
+    /// so this replaces an existing record and is a no-op once the task is
+    /// gone. The check and the write happen under the tasks lock, so a delete
+    /// cannot land between them. Use `create_task` to add a task.
     pub async fn put_task(&self, task: &ScheduledTaskRecord) -> Result<()> {
+        self.write_task(task, Write::Replace).await
+    }
+
+    async fn write_task(&self, task: &ScheduledTaskRecord, mode: Write) -> Result<()> {
         fs::create_dir_all(self.tasks_dir()).await?;
+        let root = self.root.clone();
         let path = self.task_path(&task.id);
-        write_json_file(&path, task)
-            .await
-            .with_context(|| format!("failed to write scheduled task {}", path.display()))
+        let bytes = serde_json::to_vec_pretty(task)?;
+        tokio::task::spawn_blocking(move || -> Result<()> {
+            let _lock = TasksLock::acquire(&root)?;
+            if mode == Write::Replace && !path.exists() {
+                return Ok(());
+            }
+            write_json_file_blocking(&path, &bytes)
+                .with_context(|| format!("failed to write scheduled task {}", path.display()))
+        })
+        .await?
     }
 
     pub async fn disable_task(&self, task_id: &str) -> Result<Option<ScheduledTaskRecord>> {
@@ -136,7 +155,13 @@ impl SchedulerStore {
         let Some(task) = self.get_task(task_id).await? else {
             return Ok(None);
         };
-        remove_file_if_exists(self.task_path(task_id)).await?;
+        let root = self.root.clone();
+        let path = self.task_path(task_id);
+        tokio::task::spawn_blocking(move || -> Result<()> {
+            let _lock = TasksLock::acquire(&root)?;
+            remove_file_if_exists_blocking(&path)
+        })
+        .await??;
         remove_dir_if_exists(self.runs_dir(task_id)).await?;
         Ok(Some(task))
     }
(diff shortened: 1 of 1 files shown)


crates/executor/src/scheduler_runtime.rs
```diff
@@ -117,7 +117,12 @@ pub async fn run_task(
         }
     }
     task.resume_after_fires(&plan, now_ms());
-    store.put_task(&task).await?;
+    // Guard against a delete that landed while this task was running: if
+    // the task file is gone, the task was cancelled or deleted and we
+    // must not resurrect it by writing it back.
+    if store.get_task(&task.id).await?.is_some() {
+        store.put_task(&task).await?;
+    }
```
ankrgyl: this needs to somehow be done atomically in `put_task`, otherwise, someone could delete the task between the call to `get_task` and `put_task`

CakeCrusher: agreed looking into it

CakeCrusher: Good catch — fixed in 20d83d5 by moving the guard into the store.

`update_task` writes only if the record still exists, and both it and `delete_task` hold an exclusive lock on the tasks directory, so the check and the write can't interleave with a delete. The lock is held around the file operations only, never across a run, and the OS releases it if a holder dies. `run_task` now writes back through `update_task`.

Also added a test that reproduces the original failure directly — a run holds a record, a delete lands mid-run, and the stale write-back must not recreate the task. It fails against the previous behaviour.

Related, and happy to split out if you'd prefer: the doc comment on `claim_due_tasks` notes that its read-lease-write is unconditional pending "the conditional puts in upstream PR #113". `update_task` is that primitive, so that call site could use it too. I left it alone here to keep the diff focused, and because #113 removes `claim_due_tasks` entirely — worth noting that #113 as it stands keeps the unconditional `put_task` at the end of `run_task`, so this bug survives it.

CakeCrusher: You're right, and taking it lower fixed it more cleanly than my first attempt — thanks.

Rewrote as a single store-level change: `create_task` is the only creator, so `put_task` is now replace-only and a no-op once the task is gone, with the existence check and the write under an exclusive lock on the tasks directory that `delete_task` also takes. No new method, and no call sites change — `run_task` is untouched.

Added a test that reproduces the original failure; it fails if `put_task` is allowed to create.

crates/executor/src/scheduler_store.rs
```diff
@@ -277,8 +307,44 @@ async fn write_json_file<T: Serialize>(path: &Path, value: &T) -> Result<()> {
     })
 }
 
-async fn remove_file_if_exists(path: PathBuf) -> Result<()> {
-    match fs::remove_file(&path).await {
+/// Whether a task file may be created, or only replaced if it still exists.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Write {
+    Create,
+    Replace,
```
ankrgyl: small nit is that Replace to me sounds like "if it exists, overwrite it". I spent like 10 minutes trying to understand why it wasn't doing that before reading your comment. I'd prefer names like `CreateIfNotExists` (although I think `Create` is actually `Replace`?)  and `ReplaceIfExists`

CakeCrusher: Makes sense, cleaned up the ambiguioty, also renamed to CreateOrReplace instead which is what that path actually does.

PR #297, opened 2026-09-21
Subscription auth for claude-code and codex harnesses

netojso:
## Summary

Adds first-class support for running the `claude-code` and `codex` TypeScript harnesses on a Claude Pro/Max or ChatGPT subscription instead of a per-token API key, without changing the existing API-key path.

- `Binding::Llm` gains `auth_mode: LlmAuthMode` (`api-key` default / `subscription`), exposed as `exo model register --auth-mode subscription`, which rejects `--secret`; omitting it (api-key, the default) still requires `--secret` exactly as before.
- `claudeSandboxBaseEnv` fails fast with an actionable error if subscription mode has no `CLAUDE_CODE_OAUTH_TOKEN`, and hard-fails (never silently lets the key win by precedence) if both a key and a token are present.
- The Codex sandbox boot script fails fast under subscription mode if no `auth.json` is found in `CODEX_HOME`, instead of surfacing a confusing downstream "missing bearer" error.
- Fixed a real bug found along the way: `ResponsesRuntime.fromModelBinding` built an OpenAI SDK client eagerly and crashed on a `null` apiKey, even though both harnesses only use that client for tracing (never `.complete()`). Scoped fix (`tracingOnlyModelBinding`) — harnesses that genuinely need a real key (rlm/turn-loop) are untouched.
- Added a `docker_snapshot_excludes_internal_mounted_credential_content` regression test proving `internal`-mounted credentials never leak into a Docker sandbox snapshot (structurally, not by reading Docker's docs) — also documents a more benign finding (mount *path* metadata can appear in snapshots even though content never does).
- Added `docs/subscription-auth.md`: full setup flow for both providers, the mount-write-access and mount-scope gotchas hit during verification, and the credential-hygiene guarantees/caveats.
- Every turn now logs a `subscription_auth_diagnostic` event (`{ authMode, hasKeySecret, warning }`) — never credential material.

Every slice was verified end-to-end against a real Docker sandbox and a real Claude/ChatGPT subscription on this machine, not just unit tests — see individual commit messages for the exact evidence (including a couple of real bugs found and fixed via that verification, not assumed away).

## Test plan

- [x] `pnpm check` (lint + typecheck + 170 vitest tests) green
- [x] `cargo build --workspace` / `cargo test --workspace --lib` (104 tests) green, `cargo fmt --check` clean
- [x] `docker_snapshot_excludes_internal_mounted_credential_content` run explicitly against real Docker (`--ignored`), passing
- [x] Real end-to-end turn via `--harness codex` with `--auth-mode subscription`: authenticates against ChatGPT's backend with zero `OPENAI_API_KEY` in the environment, exact expected output
- [x] Real end-to-end turn via `--harness claude-code` with `--auth-mode subscription`: authenticates via `CLAUDE_CODE_OAUTH_TOKEN` with zero `ANTHROPIC_API_KEY` in the environment, exact expected output
- [x] `exo model register` validation: subscription+secret rejected, api-key without secret rejected, both with actionable messages


netojso:
Opened against the wrong base by mistake (gh pr create defaulted to upstream instead of my fork). Re-opening against my own fork instead. Apologies for the noise.

